PR #9, opened 2020-05-21
Add onPrem explicit options and check if accessKey is set or fail

airadier:
Take the collector options out of settings and create an "onPrem" map to enable it. These options will be put inside the agent settings. This is good for MCM, where the On-Prem options are displayed in their own UI section.

An additional check is done so if the collector options appear twice (in onPrem.xxx and sysdig.settings.xxx) the helm deploy fails with an error.

Also, fail the deployment if accessKey is not specified, instead of deploying without secret or daemonset.

### DIFF
--- a/charts/sysdig/CHANGELOG.md
+++ b/charts/sysdig/CHANGELOG.md
@@ -5,6 +5,13 @@
 This file documents all notable changes to Sysdig Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## v1.7.17
+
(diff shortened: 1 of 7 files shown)


nestorsalceda (approved):
What an awesome work!

Thank you @airadier

nestorsalceda (changes requested):
Please @airadier don't forget to update the README.md documentation for onPrem

nestorsalceda (approved):

PR #12, opened 2020-06-01
[sysdig] Add an option to allow disabling captures

airadier:
## Checklist
- [x] Chart Version bumped
- [x] Title of the PR starts with chart name (e.g. `[stable/mychartname]`)

### DIFF
--- a/charts/sysdig/templates/configmap.yaml
+++ b/charts/sysdig/templates/configmap.yaml
@@ -21,6 +21,10 @@ data:
       enabled: true
 {{- end }}
 {{- end }}
+{{- if eq .Values.sysdig.disableCaptures true }}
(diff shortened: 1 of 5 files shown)


charts/sysdig/templates/configmap.yaml
```diff
@@ -21,6 +21,10 @@ data:
       enabled: true
 {{- end }}
 {{- end }}
+{{- if eq .Values.sysdig.disableCaptures true }}
+    {{- if hasKey .Values.sysdig.settings "sysdig_capture_enabled" }}{{ fail "sysdig_capture_enabled options in both .sysdig.disableCaptures and sysdig.settings.collesysdig_capture_enabledctor"}}{{ end }}
```
nestorsalceda, line 25: We are having this pattern in several places, and perhaps for next time we should think about extracting a helper function to  automate this.

nestorsalceda (approved):
LGTM! Thank you!

charts/sysdig/templates/configmap.yaml
airadier: Agree, as I already sent to PR to IBM too, I'll schedule that change for 1.7.19

PR #29, opened 2020-07-30
[sysdig-admission-controller] Use "extension-apiserver-authentication-reader" Role in kube-system instead of creating a new Role

dragon3:

This pull request updates RoleBinding  to use existing Role `extension-apiserver-authentication-reader` instead of creating a new Role.

The Role `extension-apiserver-authentication-reader` is in kube-system namespace by default, so it's better to use it instead of creating a new Role with exactly same permissions.

ref: [Set up an Extension API Server](https://kubernetes.io/docs/tasks/extend-kubernetes/setup-extension-api-server/)

> 13. Create a Kubernetes role binding from the service account in your namespace to the extension-apiserver-authentication-reader role. This allows your extension api-server to access the extension-apiserver-authentication configmap.

## Checklist
- [x] Chart Version bumped
- [ ] ~~Variables are documented in the README.md~~ (There is no new variable)


dragon3:
One more suggestion is that if we use the `extension-apiserver-authentication-reader` Role, I think it's better to rename the RoleBinding to `{{ include "sysdig-image-scanner.name" . }}-auth-reader`.

https://github.com/sysdiglabs/charts/pull/29/files#diff-1c1977d9f6d30b175afb522c8c055da2R4
```yaml
metadata:
  name: {{ include "sysdig-image-scanner.name" . }}-auth-watcher
  namespace: kube-system
```

What do you think?

airadier:
Thanks very much! I agree, probably we don't need a specific role. Let me do some tests and I'll merge. We can remove the role that we are creating in https://github.com/sysdiglabs/charts/blob/master/charts/sysdig-admission-controller/templates/role-kube-system.yaml, as it is not used anymore.

dragon3:
Thank you for the quick reply!

> probably we don't need a specific role. Let me do some tests and I'll merge. 

Thank you! let me know if it works 😁 

> We can remove the role that we are creating in https://github.com/sysdiglabs/charts/blob/master/charts/sysdig-admission-controller/templates/role-kube-system.yaml, as it is not used anymore.

Yes! This pull request will delete the manifest.

Also please check the RoleBinding naming 🙏 
`{{ include "sysdig-image-scanner.name" . }}-auth-watcher` vs `{{ include "sysdig-image-scanner.name" . }}-auth-reader`

Thank you!

airadier:
@dragon3 LGTM! I am merging in a temporary branch in order to update the CHANGELOG.md and rename the clusterrolebinding, then merging in https://github.com/sysdiglabs/charts/pull/33

Thanks very much!

dragon3:
@airadier 
Thank you!

PR #32, opened 2020-08-14
[harbor-scanner-sysdig-secure] inline scan true not yes

mstemm:
## Checklist
- [x] Chart Version bumped


nestorsalceda (approved):
Thank you for the fix @mstemm

PR #35, opened 2020-08-19
Sysdig Helm Chart output message

pabloopez:
After deployment message updated. 

To check for a successful installation of the agent, we may point users to the new Dashboard Template for Agent Health & Status, instead of Explore.

Pls @nestorsalceda review this PR. Let me know if this is not the right way to do this! Thanks!

This just updates the message displayed after deploying the agent with Helm. Or at least that is what I tried :D

## Checklist
- [ ] Chart Version bumped
- [ ] Title of the PR starts with chart name (e.g. `[mychartname]`)

None of the above fields applies I guess, it is the same message no matter the chart, no new variables to document,


nestorsalceda:
I added some fixes and merged on #36 

Thank you for the PR @pabloopez !

PR #59, opened 2020-12-15
Updated to support Agent 10.8.0 and NIA 0.1.7

KranzSysdig:

This is to update the chart for agent 10.8.0 and NIA 0.1.7. At time of the PR, agent 10.8.0 is still being baked, so please don't immediately accept. Just staging so it's ready to roll.

## Checklist
- [x] Chart Version bumped


KranzSysdig:
10.8 should be released now.

PR #68, opened 2020-12-30
[sysdig] Add templates to customize probes initial delay

jcbbc:
On some of our GKE nodes, Sysdig agents take too much time to launch and Kubernetes end-up restarting the pods in loop. By increasing the initial delay to 30s, agents have enough time to start and report successful liveness and readiness.

Current chart doesn't allow for customisation of the probes initial delay so we had to fork it and then merge new changes manually. However I don't see why this customisation would be an issue so I though it could be pushed into the official chart.

## Checklist
- [x] Chart Version bumped


nestorsalceda (approved):
LGTM

Thank you!

PR #76, opened 2021-01-22
[admission-controller] Configuration is retrieved directly from Sysdig Secure

nestorsalceda:
## Checklist
- [x] Chart Version bumped


nestorsalceda:
Forced because it tries to connect to secure backend at startup

Needs to figure out how to pass secrets here ;)

PR #106, opened 2021-03-11
ci: Test charts with secrets

tembleking:

Allows to test all the charts with secrets via environment variable.


.github/workflows/lint-test.yaml
```diff
@@ -8,24 +8,43 @@ jobs:
...
-        run: git fetch --prune --unshallow
+      - name: Set up Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.4.0
 
-      - name: Run chart-testing (lint)
-        id: lint
-        uses: helm/chart-testing-action@v1.0.0-rc.2
+      - uses: actions/setup-python@v2
```
airadier, line 19: What is using python?

airadier (approved):
I would say the big change is the upgrade from chart-testing-action@v1 to chart-testing-action@v2 :)

Some of the charts need a version bump or the linter will fail.

PR #109, opened 2021-03-18
[cloud-connector] Add extra env vars

tembleking:

Adds the extra env vars configuration to the chart values.

## Checklist
- [X] Chart Version bumped


airadier (commented):
Nice! Do we need the `extraEnvVarsCM` and `extraEnvVarsSecret`? I am ok with it but maybe overcomplicating in case no one is going to use it?

Although I see a very interesting use case, like having a shared CM with the proxy env vars and use it in multiple charts, I can't imagine many customers creating a CM, providing access to the different service accounts, etc.

airadier (approved):

PR #129, opened 2021-04-27
[sysdig] Bump agent version to 11.2.0

airadier:

Bump agent version to 11.2.0

## Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/sysdig/CHANGELOG.md
+++ b/charts/sysdig/CHANGELOG.md
@@ -9,6 +9,7 @@ This file documents all notable changes to Sysdig Helm Chart. The release number
 ### Minor changes
 
 * Update agent to 11.2.0
+* Remove --name installation parameter for `helm install` in README, unsupported in Helm 3.x
 
 ## v1.11.13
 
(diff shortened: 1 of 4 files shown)


charts/sysdig/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@
 
 This file documents all notable changes to Sysdig Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+## v1.11.14
+
+### Minor changes
+
+* Update agent to 11.2.0
+* Remove --name installation parameter for `helm install` in README, unsupported in Helm 3.x
```
balinaveen, line 12: Just to confirm ... is the --name an optional parameter in Helm < 3.0 ? Is the behavior with --name removed going to be the same as with the param switch in older versions too?

airadier: The syntax has changed from Helm 2.x to Helm 3.x. In 2.x you need to specify --name <release-name> (although I think it is mandatory to specify it) while in 3.x the release name is a mandatory parameter.

As Helm 3.x is already stable since a long time ago, and we are providing installation instructions for Helm 3.x, but some of the examples had remains of the old 2.x instructions, we are removing the --name from everywhere.

PR #136, opened 2021-04-30
[cloud-connector]: Add emptyDir for home path that allows RW

tembleking:

Allows read-write access to `/home/sysdig` through an emptyDir.

## Checklist
- [X] Chart Version bumped


airadier (approved):
LGTM, although I still don't understand the reason why STS GetCallerIdentity is failing and if making ~/.aws not read-only is the best fix.

PR #138, opened 2021-05-04
[sysdig] feat: allow specify existing or external secrets instead of creating a new one

airadier:

Allow specifying an existing secret name instead of creating a new one for storing the access-key entry. This allows using [Bitnami Sealed Secrets](https://github.com/bitnami-labs/sealed-secrets) or similar.

## Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/sysdig/values.yaml
+++ b/charts/sysdig/values.yaml
@@ -145,6 +145,8 @@ clusterName: ""
 sysdig:
   # Required: You need your Sysdig Agent access key before running agents.
   accessKey: ""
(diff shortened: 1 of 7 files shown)


charts/sysdig/values.yaml
```diff
@@ -145,6 +145,8 @@ clusterName: ""
 sysdig:
   # Required: You need your Sysdig Agent access key before running agents.
```
mmdiago, line 146: May be not describing this as required. But at least one of the following

PR #145, opened 2021-05-13
[Sysdig] Add Node Analyzer

noahkraemer:
This PR adds the Node Analyzer daemonset configmaps and helm parameters.

## Checklist
- [x] Chart Version bumped


charts/sysdig/values.yaml
```diff
@@ -247,6 +247,117 @@ nodeImageAnalyzer:
...
+      requests:
+        cpu: 150m
+        memory: 512Mi
+      limits:
+        cpu: 500m
+        memory: 1536Mi
+  hostAnalyzer:
+    image:
+      repository: sysdig/host-analyzer
+      tag: 0.1.10 # TODO pinned or latest?
```
noahkraemer: What should this tag be?

charts/sysdig/README.md
```diff
@@ -43,70 +43,107 @@ The command removes all the Kubernetes components associated with the chart and
...
+| `nodeAnalyzer.imageAnalyzer.resources.requests.memory`     | Node Image Analyzer Memory requests per node                                             | `512Mi`                                                                       |
+| `nodeAnalyzer.imageAnalyzer.resources.limits.cpu`          | Node Image Analyzer CPU limit per node                                                   | `500m`                                                                        |
+| `nodeAnalyzer.imageAnalyzer.resources.limits.memory`       | Node Image Analyzer Memory limit per node                                                | `1536Mi`                                                                      |
+| `nodeAnalyzer.hostAnalyzer.image.repository`               | The image repository to pull the Host Analyzer from                                      | `sysdig/host-analyzer`                                                        |
+| `nodeAnalyzer.hostAnalyzer.image.tag`                      | The image tag to pull the Host Analyzer                                                  | `TODO 0.1.7`                                                                  |
```
noahkraemer: What should this tag be?

charts/sysdig/README.md
```diff
@@ -125,7 +162,7 @@ $ helm install --namespace sysdig-agent sysdig-agent -f values.yaml sysdig/sysdi
 > **Tip**: You can use the default [values.yaml](values.yaml)
 
 ## Node Image Analyzer
-
+ TODO Update
```
noahkraemer: We will still default to installing the NIA. Should we change the default to the NA? Should we remove the NIA entirely?

charts/sysdig/values.yaml
nestorsalceda: We use pinned versions following the convention: 

https://helm.sh/docs/chart_best_practices/pods/#images

Although it involves a bit of more work in our side, I would keep doing it.

charts/sysdig/README.md
nestorsalceda: Same than in the previous

charts/sysdig/README.md
nestorsalceda: I think we should promote the NA, perhaps we can add a major version bump to make this clear.

charts/sysdig/values.yaml
noahkraemer: Agreed =) I just need the latest version so I can use that

charts/sysdig/README.md
noahkraemer: Sounds good to me 👍

charts/sysdig/values.yaml
nestorsalceda: The CI failed, I think is because the comment

Once you remove it, it will be fine

nestorsalceda:
Thank you so much @nkraemer-sysdig !!

PR #164, opened 2021-05-26
[Sysdig] host analyzer: add hostBase

luca-sd:
By specifying the host base we can ensure that host scanning will scan the host and not the container it's in.

## Checklist
- [x] Chart Version bumped


charts/sysdig/values.yaml
```diff
@@ -332,6 +332,8 @@ nodeAnalyzer:
     # The list of directories to inspect during the scan, expressed as a comma separated list.
     # dirsToScan: "/etc,/var/lib/dpkg,/usr/local,/usr/lib/sysimage/rpm,/var/lib/rpm,/lib/apk/db"
 
+    hostBase: "/host"
```
airadier, line 335: Are we sure we need to expose `hostBase` as an option in values? I see the mountPath for root-vol is hardcoded to /host in https://github.com/luca-sd/charts/blob/45e7278014cd734dedb4c7e8b4cb6a493a734d81/charts/sysdig/templates/daemonset-node-analyzer.yaml#L214, so wouldn't it make sense to just hardcode this value in the configmap?

luca-sd: It makes sense but I've got a question: can you see issues in case someone used a configmap that has this environment variable updating with a daemonset that doesn't have it exposed anymore?

airadier:
@luca-sd don't go crazy with the "Error: [trailing-spaces] trailing spaces" error. There is a flaw in our workflow that causes this problem when a PR comes from an external fork. Let me clone your PR so it triggers the check correctly :)

luca-sd:
Thanks Alvaro, anyways I like the daemonset solution better.

airadier:
PR https://github.com/sysdiglabs/charts/pull/166 includes the same changes you are proposing in https://github.com/draios/sysdig-cloud-scripts/pull/158. I think we can close this PR. Is it ok?

luca-sd:
We have decided to use the DS solution, bundle it up with another NA fix in https://github.com/sysdiglabs/charts/pull/166 and port it to the regular DS as well ( https://github.com/draios/sysdig-cloud-scripts/pull/158 ) so I'm closing this one

PR #169, opened 2021-06-07
[sysdig] Fix eBPF+slim on GKE

gnosek:
We need /etc/os-release mounted to detect Google COS

## Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version bumped
- [x] PR only contains changes for one chart

PR #170, opened 2021-06-07
[sysdig] Fix eBPF+slim on GKE

airadier:
Clone https://github.com/sysdiglabs/charts/pull/169 for CI


airadier:
It would be perfect if you can also update the CHANGELOG.md file 🙏 

Thanks very much!

PR #184, opened 2021-07-08
[sysdig] Prefer .Chart.AppVersion over .Values.image.tag

deepskyblue86:
Use Chart.yaml appVersion as the default image tag.
This comes in handy when bumping the version, other than being a good practice.

## Checklist


airadier:
Hi @deepskyblue86, thanks for the PR I agree with the good practice and we use appVersion in other charts.

However I have a concern: this chart is deploying multiple components. The agent is the central piece, but we also have the node image analyzer, host analyzer and benchmark runner. All the image tags are now specified in the values.yaml

Should we use the chart appVersion for the agent, allow to override via values (and keep the entry in the values.yaml, empty, with a comment referencing appVersion in chart.yaml), and use the values for the tags of the other components?

@balinaveen @nestorsalceda what's your opinion on this?

deepskyblue86:
@airadier my change was actually driven by the need of being able to easily automate the version change.
Your point is perfectly valid. Most probably better to come out with another solution.

balinaveen:
> @balinaveen @nestorsalceda what's your opinion on this?
This is a good question. @deepskyblue86  will schedule a call to discuss it.

PR #186, opened 2021-07-13
[sysdig] Version bump scripts

deepskyblue86:
## Checklist

### DIFF
--- a/scripts/chart-version-bump.sh
+++ b/scripts/chart-version-bump.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env sh
+
+set -e
+
+awk '
+{
+    if ($1 == "version:")
+    {
+        l = split($NF, v, ".");
+        v[l]++;
+        for (i=1; i<l; ++i)
+            V = V v[i] "."
+        V = V v[i]
+        $NF=V
+    }
+    print
+}' Chart.yaml > Chart.yaml.2
+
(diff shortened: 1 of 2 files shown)


airadier:
Thanks!

You'll need to either bump the chart.yaml version (or the linter will fail) or move the scripts to a different location outside the charts/folder.

Do we want to have these scripts shipped in the helm chart package? Maybe we could store them in /scripts at the root of the repository instead of the chart itself?

Otherwise LGTM.

balinaveen:
> Do we want to have these scripts shipped in the helm chart package? Maybe we could store them in /scripts at the root of the repository instead of the chart itself?
> 
I agree it doesn't make sense to ship these scripts as part of the helm chart package. It's for internal use, so moving them up in the directory is a good idea.

deepskyblue86:
I didn't know they'd be shipped this way, that wasn't intentional. I'm moving them.

airadier:
> I didn't know they'd be shipped this way, that wasn't intentional. I'm moving them.

No problem! Yes, they won't break anything in the chart, but the full folder is packaged as a .tar.gz file which is pulled by Helm when installing the chart. So it makes more sense to maintain that scripts in the /scripts folder outside of the chart, as per the latest version.

Thanks!

airadier (approved):

deepskyblue86:
Please mind I don't have merge grants :upside_down_face:

PR #203, opened 2021-08-04
[admission-controller] SSPROD-8596 add support for dockercfg in inline-scan-service

airadier:

Add an option to specify a dockerCfg secret name. This is automatically created in Openshift and provides pull permissions from the internal registry. Setting the name will mount the secret inside the inline-scan-service container and add an option so --registry-auth-dockercfg flag is added to the inline-scan script.

## Checklist
- [x] Chart Version bumped


nestorsalceda (approved):
Thank you @airadier !!

PR #205, opened 2021-08-06
[sysdig] Change the default agent container resources

deepskyblue86:
Default agent container resources are too low, change them to be good for a fair server.
## Checklist
- [x] Chart Version bumped


balinaveen (approved):
LGTM.

charts/sysdig/values.yaml
```diff
@@ -26,11 +26,11 @@ resources:
   # a sane defaults ones. If you have more questions or concerns, please refer
   # to Sysdig Support for more info about it
```
omerazr, line 27: Can we please point to Sysdig docs instead of asking them to approach support?

charts/sysdig/values.yaml
```diff
@@ -26,11 +26,11 @@ resources:
...
   requests:
-    cpu: 600m
-    memory: 512Mi
+    cpu: 3000m
```
omerazr, line 29: This can't be the default value. I might be missing something, but it's way above our advertised number.

charts/sysdig/values.yaml
```diff
@@ -26,11 +26,11 @@ resources:
...
+    memory: 3072Mi
   limits:
-    cpu: 2000m
-    memory: 1536Mi
+    cpu: 3000m
```
omerazr, line 32: why are we setting the limits and requests to be the same number

charts/sysdig/values.yaml
balinaveen: It matches our "mid sized" node config (~32 core nodes on ~100 node clusters)

charts/sysdig/values.yaml
balinaveen: To ensure that the agent has sufficient resources on oversubscribed nodes. It is precisely when the load on a node  approaches it's total available value that the agent's resource usage will go above requests and approach limits but if requests are considerably less than limits, then the agent will likely not get much more than requests in this scenario.

PR #209, opened 2021-08-10
Add /sys/kernel/debug VolumeMount to helm chart, when eBPF is in use

jcpittman144:
Agent eBPF code requires access to the /sys/kernel/debug filesystem.
For certain kernel versions, this filesystem is not accessible by default
from within docker containers, so add a VolumeMount in the helm chart.

## Checklist
- [x ] PR only contains changes for one chart


charts/sysdig/templates/daemonset.yaml
```diff
@@ -229,6 +233,9 @@ spec:
         {{- if (and .Values.ebpf.enabled .Values.slim.enabled) }}
         - name: bpf-probes
           emptyDir: {}
+	- name: sys-tracing
+          hostPath:
+	    path: /sys/kernel/debug
```
airadier, line 238: It looks there is some weird indentation here? Maybe tabs vs spaces?

airadier (changes requested):

airadier:
Thanks for the PR! The linter is failing because chart version was not updated un chart.yaml. Could you please update that and rhe CHANGELOG.md file? I noticed some identation issues that might need fixing too.

charts/sysdig/templates/daemonset.yaml
jcpittman144: Thanks for the review.  I will fix the indentation, update the version and CHANGELOG, and resubmit the PR.
Thanks!

jcpittman144:
Closing this PR, rebasing against an updated version of the repo.

PR #213, opened 2021-08-11
Add /sys/kernel/debug VolumeMount to helm chart, when eBPF is in use

jcpittman144:
Agent eBPF code requires access to the /sys/kernel/debug filesystem.
For certain kernel versions, this filesystem is not accessible by default
from within docker containers, so add a VolumeMount in the helm chart.

## Checklist
- [x] Chart Version bumped


airadier (approved):
LGTM

PR #214, opened 2021-08-12
[admission-controller] Add https proxy management

miketnt:
## Checklist
- [x] Chart Version bumped


charts/admission-controller/values.yaml
```diff
@@ -101,6 +103,8 @@ scanner:
   # Put your <Sysdig-Secure-Secret-Token> in this value
   # Set to http://proxyIp:proxyPort in case connection to Sysdig Secure requires a proxy
   httpProxy: ""
+  # Set to https://proxyIp:proxyPort in case connection to Sysdig Secure requires a proxy
```
airadier: Better use plain http:// in the example...

airadier (changes requested):
You'll need to update the version in Chart.yaml to pass the linter check.

charts/admission-controller/values.yaml
miketnt: Updated both the example and the version in the Chart. We're on the road again. Thanks @airadier!

airadier (approved):

airadier:
LGTM

PR #216, opened 2021-08-18
[admission_controller] add option for existing secret

endriu0:
This is my attempt at adding existing secrets to admission-controller. Once merged will hopefully allow moving away from storing the secrets in clear texts in the values and allow to use things like externalsecrets.

## Checklist
- [x] Chart Version bumped


endriu0:
So don't want to say 'It works on my machine' but that lint seems to pass locally - not sure why it fails here...

PR #219, opened 2021-08-24
[sysdig] Pass namespace name and and pod name to the agent container

fremmi:

With this PR the agent will be able to read its namespace and pod name from the local filesystem
This is useful in the leader election mechanism

## Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/sysdig/Chart.yaml
+++ b/charts/sysdig/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: sysdig
-version: 1.12.13
+version: 1.12.14
 description: Sysdig Monitor and Secure agent
 keywords:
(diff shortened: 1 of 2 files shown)


charts/sysdig/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: sysdig
-version: 1.12.12
-appVersion: 11.3.0
+version: 1.12.13
```
deepskyblue86: as discussed, now you need to rebase and bump it up

deepskyblue86 (changes requested):
Please update charts/sysdig/CHANGELOG.md too

PR #227, opened 2021-09-01
[sysdig] Add podinfo volume

bawhetst:
Adds a podinfo volume that uses the downward API to expose the pod name and namespace in the container. This is used by the 12.0.0 agent to autodetect the agent namespace.

## Checklist
- [x] Chart Version bumped


bryanseay (changes requested):
~Change looks good. Marking as "Request changes" to block merging until 12.0.0 is released.~
That didn't work.

fremmi (approved):

PR #229, opened 2021-09-08
docs: remove sysdig/sysdig instructions from gh-pages

deepskyblue86:
Avoid having sysdig/sysdig instructions in gh-pages


luca-sd (approved):
lgtm

PR #232, opened 2021-09-09
[sysdig] fix: restore appVersion

deepskyblue86:
People like to have appVersion, even if optional. Restore it and enhance the tooling.

## Checklist
- [x] Chart Version bumped

### DIFF
--- a/scripts/sysdig/image-version-bump.sh
+++ b/scripts/sysdig/image-version-bump.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 set -e
 
--- a/scripts/sysdig/image-version-bump.sh
+++ b/scripts/sysdig/image-version-bump.sh
@@ -39,5 +39,20 @@ BEGIN {
     print
 }
 ' values.yaml > values.yaml.2
-
 mv values.yaml.2 values.yaml
+
+
+awk $@ '
+BEGIN {
+    if (!AGENT_VERSION)
+        exit 1
+}
+
+{
+    if ($1 ~ /^appVersion:/)
+        sub(/:.*/, ": "AGENT_VERSION)
+
+    print
+}
+' Chart.yaml > Chart.yaml.2
+mv Chart.yaml.2 Chart.yaml
--- a/charts/sysdig/Chart.yaml
+++ b/charts/sysdig/Chart.yaml
@@ -1,6 +1,7 @@
 apiVersion: v1
 name: sysdig
 version: 1.12.16
+appVersion: 11.4.1
 description: Sysdig Monitor and Secure agent
 keywords:
   - monitoring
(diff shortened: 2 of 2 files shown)


scripts/sysdig/image-version-bump.sh
```diff
@@ -39,5 +39,20 @@ BEGIN {
...
+}
+
+{
+    if ($1 ~ /^appVersion:/)
+        sub(/:.*/, ": "AGENT_VERSION)
+
+    print
+}
+' Chart.yaml > Chart.yaml.2
+mv Chart.yaml.2 Chart.yaml
```
irozzo-1A, line 58: Personally I would consider more natural to use `sed` here, it would be even more concise e.g.
```
sed -i -e "s/\^(appVersion:\).*/\1 ${AGENT_VERSION}/g"
```
what do you think? (syntax to be checked, I did not run it)

deepskyblue86: Me too, but the script arguments are designed to be passthrough to awk, so we don't have `${AGENT_VERSION}` here, we have e.g. `-v AGENT_VERSION=1.2.3` and I wouldn't bother to parse `$@`.
I considered taking it like this:
```
AGENT_VERSION=$(echo "$@" | grep -Po '(?<=AGENT_VERSION=)[^ ]+')
```
but I preferred to keep the same style as the first part of the script.

irozzo-1A: I see, I would have probably used sed everywhere, but it's largely matter of taste. LGTM ;-)

deepskyblue86: As far as I know sed doesn't work on multiple lines, I should have used perl but I can't.
Maybe you didn't notice, but the format is:
```$ grep -A 1 repository values.yaml :
values.yaml:  repository: sysdig/agent
values.yaml-  tag: 11.4.1
--
values.yaml:    repository: sysdig/agent-kmodule
values.yaml-
--
values.yaml:    repository: sysdig/node-image-analyzer
values.yaml-    tag: 0.1.10
--
values.yaml:      repository: sysdig/node-image-analyzer
values.yaml-      tag: 0.1.12
--
values.yaml:      repository: sysdig/host-analyzer
values.yaml-      tag: 0.1.3
--
values.yaml:      repository: sysdig/compliance-benchmark-runner
values.yaml-      tag: 1.0.6.0
```
and the awk it's reading the "repository" line and then doing `getline`.

irozzo-1A: yep, I saw that. I think it should be possible with `sed`
e.g. this seems to do something similar:
https://github.com/cilium/cilium/blob/d8334ae9134a0905ea562fc0ca193dea9384487d/install/kubernetes/Makefile#L68-L81

deepskyblue86: I've never seen such a syntax before. That's amazing.
Anyway, for now I prefer to keep it like this, and coordinate with the user of the script to change also the way arguments are passed.

irozzo-1A: sure, no pb. it was just for reference ;-)

PR #250, opened 2021-09-27
[sysdig] update changelog to notify of breaking change

allethom:
## Notify client of breaking change
In chart version 1.12.13, the default agent container resources was set to [small](https://docs.sysdig.com/en/tuning-sysdig-agent.html). This was a breaking change (and not minor as originally stated) because upgrading the agent using this chart from an earlier chart where the default was [medium](https://docs.sysdig.com/en/tuning-sysdig-agent.html) could result in less resources configured for it than required.

## Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/sysdig/CHANGELOG.md
+++ b/charts/sysdig/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 This file documents all notable changes to Sysdig Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+## v1.12.23
+
+### Breaking change notification
+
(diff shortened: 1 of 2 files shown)


charts/sysdig/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 This file documents all notable changes to Sysdig Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+## v1.12.23
+
+### Breaking change notification
+
+- In chart version 1.12.13, the default agent container resources was set to [small](https://docs.sysdig.com/en/tuning-sysdig-agent.html). A breaking change was also introduced in the same chart version, where upgrading agent can result in the wrong amount of resources used by the agent.
```
balinaveen: Suggest rewording this as follows
"In chart version 1.12.13, the default agent container resources was set to [small] (https://docs.sysdig.com/en/tuning-sysdig-agent.html). This was a breaking change (and not minor as originally stated) because upgrading the agent using this chart from an earlier chart where the default was [medium] could result in less resources configured for it than required."

allethom: updated to reflect different wording

deepskyblue86 (approved):

PR #264, opened 2021-10-14
[sysdig] Allow agent-slim image repo to be configured in full

achandras:

We can now specify the slim agent, rather than assuming that it will always be suffixed with `-slim`.

## Checklist
- [x] Chart Version bumped


alblasco (approved):
LGTM

achandras:
Using https://github.com/sysdiglabs/charts/pull/277 instead

PR #273, opened 2021-10-27
[sysdig] Version bump scripts update

aq17:
Related to https://sysdig.atlassian.net/browse/SSPROD-9633, currently the script fails when not bumping the agent version, but we want to implement this for the benchmark runner as well

## Checklist


airadier (approved):
Hi @aq17 ! Really I am not familiar with this script, so adding @deepskyblue86 for review. I notice that we are already checking that either AGENT_VERSION or IMAGE_ANALYZER_VERSION or HOST_ANALYZER_VERSION is checked at the first was command. So I understand this change will make the second part (updating the app version in Chart.yaml) completely optional, right?

LGTM!

PR #280, opened 2021-10-28
fix(cloud-connector, doc): expose image.tag

wideawakening:
For internal testing, I use quite a lot the combo `image.repository` +  `image.tag` for overriding the chart, but the latest is not exposed in the README / docs.

## Checklist


charts/cloud-connector/README.md
```diff
@@ -21,6 +21,7 @@ chart and their default values:
 | -------------------------------- | --------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------- |
 | `replicaCount`                   | Amount of replicas for Cloud Connector                                                                          | `1`                                                                             |
 | `image.repository`               | The image repository to pull from                                                                               | `sysdiglabs/cloud-connector`                                                    |
+| `image.tag`               	    | The image tag	                                                                                                 |                                                                                 |
```
wideawakening: didn't want to put any default value, as it's not using `latest` perse, but the pinned-version of the latest image on the helm hub.
I see the explicit version on  the `Chart.yaml` file but don't wanna add maintenance mismatch on this file

wideawakening: checked other internal charts and only one exposing that var is this one, and it's pinned
https://github.com/sysdiglabs/charts/blob/master/charts/sysdig-stackdriver-bridge/README.md

outside sysdiglabs, not many charst on the arifactoryhub expose that neat values table like we do;)
but found some examples and they pin it too

https://artifacthub.io/packages/helm/bitnami/external-dns
image

https://artifacthub.io/packages/helm/elastic/elasticsearch
image

https://artifacthub.io/packages/helm/bitnami/mysql
image

wideawakening: agreed with @nestorsalceda to add it, with the description annotation of it being immutable  + not to pin it but say `latest pinned version` to lower maintenance cost

PR #286, opened 2021-11-03
[sysdig] Removed invalid image pull secret reference from initContainers

jbehrends:
Helm: v3.5.3
K8s: 1.19.13

When trying to install the latest version of the helm module using the "slim" container option you get the following error:
**Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(DaemonSet.spec.template.spec.initContainers[0]): unknown field "imagePullSecrets" in io.k8s.api.core.v1.Container**

After debugging and doing some research, it appears that imagePullSecrets is not part of the containers or initcontainers objects.. 
https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#containers

In testing, after removing imagePullSecrets from under "initContainers" the Chat was able to be validated, and installed correctly with a private registry using an image pull secret that was already defined higher in the manifest.

## Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/sysdig/CHANGELOG.md
+++ b/charts/sysdig/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 This file documents all notable changes to Sysdig Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+## v1.12.35
+### Minor changes
+
+- Removed invalid imagePullSecrets from initContainers
+
 ## v1.12.31
 
 ### Minor changes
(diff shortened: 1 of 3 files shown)


airadier:
Looks good to me. The imagePullSecrets are already at the spec level, so having them inside a container is a mistake and invalid, as you already pointed.

achandras (approved):
LGTM, thanks for this fix!

PR #316, opened 2022-01-03
chore: Update lint action to use pull_request_target

achandras:

This is one possible solution to the issue of PRs from forks not being able to access secrets. There is a potential [security risk](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) to this event, but other than helm charts, nothing is really built and run that could pose a threat.

Alternatively, we could use completely fake values for the tokens in `ci/*.yaml.template`, which works for the sysdig chart, but I have not tested for the other charts.

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -1,6 +1,6 @@
 name: Lint and Test Charts
 
-on: pull_request
+on: pull_request_target
 
 jobs:
   lint-test:
@@ -10,6 +10,8 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
 
       - name: Set up Helm
         uses: azure/setup-helm@v1


airadier:
Hi @achandras , so if I get it correctly, the pull_request_target event will execute anytime a change in the pull request happens, but the context is the base branch (usually 'master') instead of the merge commit, right? Then, with:

```
          ref: ${{github.event.pull_request.head.ref}}
          repository: ${{github.event.pull_request.head.repo.full_name}}
```

we checkout the head of the forked (in case it is a fork) repository.

So:
- It will run the action YAMLs from the base branch (master)
- It will checkout and execute the actions on the pull request branch

If it works as expected, it seems safer, as it won't allow running modified YAML for the workflow actions, but I still see some security risks. For example, you could automatically trigger the execution of a modified helm chart that is configured using our secrets and then runs a custom image in the kind cluster that sends the values of the secrets somewhere. And the execution would be automatic.

I wonder if an approach similar to https://github.com/imjohnbo/ok-to-test would be safer. I like the ok-to-test action, but an alternative could be: 

- Make the linting and testing action trigger on a "workflow_call" event, not on pull_request or pull_request_target
- On pull_request event, check if it is triggering from **our repo (not a fork)**. If it is, then trigger the linting workflow (see  https://docs.github.com/en/actions/learn-github-actions/reusing-workflows)
- For pull requests coming from a fork, we can have another workflow that is triggered by "workflow_dispatch" (manually), so in case we need to execute the linting on a PR from a forked repo, we can trigger this workflow, which internally would just reuse the linting and testing workflow.

achandras:
Yes, while it is safe from the threat of leaking repo write credentials, you're right that the access tokens can still be leaked on install by someone determined. Even a modified `NOTES.txt` would be sufficient 😄

Which means that some portion testing for forks will always have to be manually triggered, and I like the `ok-to-test` workflow for that. Happy to update

PR #329, opened 2022-01-20
[sysdig] - Add Image Digest Options

KReg27:
Allows for the addition of digest values to image pulls to meet regulatory requirements.

## Checklist
- [x] Chart Version bumped


deepskyblue86:
@KReg27 thanks for this. Would you mind rebasing?

achandras:
@KReg27 can we use only the digest when it's available, rather than tag and digest together? Since the digest will be given preference anyway, is there any benefit to keeping both?

charts/sysdig/templates/_helpers.tpl
```diff
@@ -68,7 +68,7 @@ Return the proper Sysdig Agent image name
 {{- if .Values.image.overrideValue }}
     {{- printf .Values.image.overrideValue -}}
 {{- else -}}
-    {{- include "sysdig.imageRegistry" . -}} / {{- include "sysdig.repositoryName" . -}} : {{- .Values.image.tag -}}
+    {{- include "sysdig.imageRegistry" . -}} / {{- include "sysdig.repositoryName" . -}} : {{- .Values.image.tag -}} {{- if .Values.image.digest -}} @ {{- .Values.image.digest -}} {{- end -}}
```
achandras: @KReg27 can we use only the digest when it's available, rather than tag and digest together? Since the digest will be given preference anyway, is there any benefit to keeping both?

KReg27: @achandras No specific reason. I have made the change as suggested.

achandras:
@KReg27 thanks for the updates! Last request, do you mind updating the version in `Chart.yaml` and moving the changelog section to `v1.12.46`?

KReg27:
> @KReg27 thanks for the updates! Last request, do you mind updating the version in `Chart.yaml` and moving the changelog section to `v1.12.46`?

Done!

achandras:
Thanks!

PR #339, opened 2022-02-04
[sysdig] trim whitespace around else clause for image tags

achandras:

Fixes issue where there is whitespace separating a tag from the image.

## Checklist
- [x] Chart Version bumped


balinaveen (approved):
I'm not completely familiar with the syntax but it seems to be consistent with how '- 'is used elsewhere, so I will approve.

PR #349, opened 2022-02-24
[sysdig] update for tech preview release of Runtime Scanner

ProvoK:


charts/sysdig/values.yaml
```diff
@@ -474,11 +476,11 @@ nodeAnalyzer:
     resources:
       requests:
         cpu: 150m
-        memory: 512Mi
+        memory: 1024Mi
```
blkt: Given that this is a DaemonSet, the option (1) of increasing requests could prevent scheduling of the pod if requests are unsatisfiable, while option (2) of having lower requests and high limits (i.e. `512Mi/2Gi`) guarantees the `helm install ...` command terminates and possibly only at a later time have problems finding enough memory to run.

Are we fine going with option (1)?

blkt (approved):
LGTM

charts/sysdig/values.yaml
ProvoK: It's likely not gonna work - as of now - with 512Mi, improvements for memory usage are in the pipeline though.

But I see the point in preferring a single container OOMKilled than the entire Pod "Pending".
I'm rollbacking this change

ProvoK: (done)

blkt (approved):

PR #359, opened 2022-03-11
[agent] standalone Sysdig agent

achandras:

This is a chart that splits off the agent components of the sysdig chart. It is expected that there will be some drift until the full switchover.

There is not much to review, as most of this is a copy with some cleanup. I have called out newly added stuff if you would like to take a look at it.

## Checklist


charts/agent/templates/_helpers.tpl
```diff
@@ -0,0 +1,194 @@
...
+*/}}
+{{- define "get_or_fail_if_in_settings" -}}
+{{- $keyValue := tpl (printf "{{- .Values.%s -}}" .key) .root }}
+{{- if $keyValue -}}
+    {{- if hasKey .root.Values.sysdig.settings .setting }}{{ fail (printf "Value '%s' is also set via .sysdig.settings.%s'." .key .setting) }}{{- end -}}
+    {{- $keyValue -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
```
achandras, line 152: The following 3 helpers are just to allow globals as a fallback

charts/agent/templates/_helpers.tpl
```diff
@@ -0,0 +1,194 @@
...
+    */}}
+    {{- .Values.sysdig.existingAccessKeySecret | default .Values.global.sysdig.accessKeySecret | default "" -}}
+{{- end -}}
+
+{{- define "agent.clusterName" -}}
+    {{- $clusterName := include "get_or_fail_if_in_settings" (dict "root" . "key" "clusterName" "setting" "k8s_cluster_name") }}
+    {{- $clusterName | default .Values.global.clusterConfig.name | default "" -}}
+{{- end -}}
+
+{{/*
```
achandras, line 174: This is also a new set of helpers to determine when to automatically use ebpf.

airadier: This is nice. I don't know how well it plays with "helm template" (i.e. some customers generate the YAMLs and then apply on the cluster). Does it work offline?

airadier (approved):
Looks good. My only concern, addressed in an inline comment, is if the node COS autodetection will work for "offline" use cases, like when using `helm template`.

My feeling about EBPF is, in the end this should be exposed as a configuration option, but it should be detected by the agent automatically, so something like:

- Try precompiled kernel probe
- Try building probe
- Fallback to EBPF

should belong to the logic inside the agent startup, and not to the helm chart. Is there any technical limitation preventing the fallback to EBPF behavior in the agent?

achandras:
@airadier that's a good callout - currently `lookup` doesn't work with `helm template`, even with `--validate`/`--dry-run`. But in that case, the `ebpf.enabled` flag will still work as today.

I think fallback to ebpf could be something added once it's officially supported, since from the [docs](https://docs.sysdig.com/en/docs/installation/sysdig-agent/agent-installation/agent-install-kubernetes/steps-for-gke/#gke-cosebpf-specific-requirements) it seems like we only support it for GKE COS

PR #360, opened 2022-03-11
[sysdig-deploy] initial sysdig-deploy chart

achandras:

Adds the initial stub of the sysdig-deploy umbrella chart. Currently it only has the agent chart as a dependency.

Depends on #359 first

## Checklist


achandras:
Merging since this is just a stub

PR #366, opened 2022-03-16
[sysdig-agent] Include the configmap for new runtime scanner

kadkins-sysdig:
The configmap for the new runtime scanner needs to be mounted in the daemonset for it to function.

## Checklist
- [ ] Chart Version bumped
- [ x] Variables are documented in the README.md
- [ x] Title of the PR starts with chart name (e.g. `[mychartname]`)
- [ x] PR only contains changes for one chart


kadkins-sysdig:
It turns out the issue I was encountering will not be fixed by this PR. I ended up adding more disk space and it looks like the issue may be resolved. I am withdrawing this PR.

PR #367, opened 2022-03-17
[runtime-scanner]: Release version 0.2.3 with CRIO support

dippie8:
Mount volumes for containers-storage data to allow the `runtimescanner` to load image data from the cri-o clusters.
## Checklist


ProvoK (approved):
LGTM thanks @dippie8

PR #369, opened 2022-03-17
Include values.yaml in gh-pages

airadier:
Improve release process to include values.yaml, as some README.md are linking to them using relative paths.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,5 +35,6 @@ jobs:
           cp -r charts ${tmp_folder}/
           pushd ${tmp_folder}
           git add -A "charts/**/README.md"
-          git pull && git commit -m "Updating README.md files" && git push
+          git add -A "charts/**/values.yaml"
+          git pull && git commit -m "Updating README.md and values files" && git push
           popd


wideawakening:
> Improve release process to include values.yaml, as some README.md are linking to them using relative paths.

what would be an actual example of README with relative path failing to pin values?

airadier:
> > Improve release process to include values.yaml, as some README.md are linking to them using relative paths.
> 
> what would be an actual example of README with relative path failing to pin values?

- The agent one: https://charts.sysdig.com/charts/agent/ as reported in https://sysdigcloud.slack.com/archives/C80UNTWBD/p1647533958413729.
- Admission controller: https://charts.sysdig.com/charts/admission-controller/
- Sysdig stackdriver bridge: https://charts.sysdig.com/charts/sysdig-stackdriver-bridge/

I fixed temporarily via[ manually adding the values.yaml to gh-pages](https://github.com/sysdiglabs/charts/commit/672b06879361797e2e9dc710a54bf93512594246), and this is the automated step on every release.

Alternatively, we could fix the README.mds to use absolute links to the values.yaml instead.

wideawakening:
image
oh i see :)

because repo is public can also leave a link to the specific file with a link, like we have in admission
image

that way can see it without downloading, but whatever suits 👍🏽

PR #375, opened 2022-03-21
[sysdig] add env for kmodule initContainer

achandras:

Adds `slim.env` for passing in environment variables to the kernel module builder init container.

## Checklist
- [x] Chart Version bumped


charts/sysdig/README.md
```diff
@@ -109,6 +109,7 @@ The following table lists the configurable parameters of the Sysdig chart and th
 | `slim.resources.requests.cpu`                                        | CPU requested for building the kernel module                                             | `1000m`                                                                        |
 | `slim.resources.requests.memory`                                     | Memory requested for building the kernel module                                          | `348Mi`                                                                        |
 | `slim.resources.limits.memory`                                       | Memory limit for building the kernel module                                              | `512Mi`                                                                        |
+| `slim.env`                                                           | Environment variables for the kernel module image builder. Provide as map of `VAR: val`  | `{}`                                                                           |
```
deepskyblue86: Since it's specific to kmodule, I'd like it better to have the name reflect it, otherwise one could expect them to be applied to the slim container too. Or else we could just actually apply them to both.

deepskyblue86 (commented):
Shouldn't we have the same changes in `charts/agent`?

achandras:
@deepskyblue86 the changes will be pulled in with a separate PR

charts/sysdig/README.md
achandras: updated to `daemonset.kmodule.env`, thanks @alblasco for the suggestion

charts/sysdig/templates/daemonset.yaml
```diff
@@ -98,6 +98,10 @@ spec:
             - name: SYSDIG_BPF_PROBE
               value:
           {{- end }}
+          {{- range $key, $value := .Values.slim.env }}
```
deepskyblue86:
```suggestion
          {{- range $key, $value := .Values.daemonset.kmodule.env }}
```

achandras: 🤦 doh... thanks for catching this

deepskyblue86 (approved):

PR #376, opened 2022-03-25
[sysdig] Fast rolling update

lilx1ao:

fasten up the Kubernetes nodes rolling update process

## Checklist
- [x] Chart Version bumped


charts/sysdig/templates/daemonset.yaml
```diff
@@ -303,5 +303,7 @@ spec:
                 fieldPath: metadata.name
               path: name
   updateStrategy:
-{{ toYaml .Values.daemonset.updateStrategy | indent 4 }}
+    type: RollingUpdate
+    rollingUpdate: 
+      maxUnavailable: {{include "sysdig.parallelStarts" .}}
```
lilx1ao, line 309: I'm assuming the customers won't specify the maxSurge value

charts/sysdig/templates/_helpers.tpl
```diff
@@ -231,3 +231,18 @@ Sysdig Eve Connector Secret generation (if not exists)
 {{ randAlphaNum 32 | b64enc | quote }}
 {{- end -}}
 {{- end -}}
+
+{{/*
+to help the maxUnavailable and max_parallel_cold_starts pick a reasonable value based on the cluster size
+*/}}
+{{- define "sysdig.parallelStarts" -}}
+{{- if .Values.daemonset.updateStrategy.rollingUpdate -}}
+    {{- .Values.daemonset.updateStrategy.type.rollingUpdate.maxUnavailable -}}
```
lilx1ao: use the specified maxUnavailable in values.yaml

charts/sysdig/templates/configmap.yaml
```diff
@@ -46,6 +46,8 @@ data:
 {{- if $sslVerifyCertificate }}
     ssl_verify_certificate: {{ $sslVerifyCertificate }}
 {{- end }}
+    k8s_coldstart:
+      max_parallel_cold_starts: {{ include "sysdig.parallelStarts" .}}
```
lilx1ao, line 51: if max_parallel_cold_starts is specified in values.yaml, this will be overwritten by the template in line 52

charts/sysdig/templates/daemonset.yaml
achandras: I think having it there is a good idea, since it's only these two values in the [spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#rollingupdatedaemonset-v1-apps) at the moment.

something like `maxSurge: {{ .Values.daemonset.updateStrategy.rollingUpdate.maxSurge }}`

charts/sysdig/templates/configmap.yaml
```diff
@@ -46,6 +46,8 @@ data:
 {{- if $sslVerifyCertificate }}
     ssl_verify_certificate: {{ $sslVerifyCertificate }}
 {{- end }}
+    k8s_coldstart:
```
achandras, line 50: let's add a check around this, so that this only gets added if someone hasn't defined `k8s_coldstart` in the `sysdig.settings`. while some yaml parsers will work around duplicate keys, it's really invalid

charts/sysdig/templates/daemonset.yaml
```diff
@@ -303,5 +303,7 @@ spec:
                 fieldPath: metadata.name
               path: name
   updateStrategy:
-{{ toYaml .Values.daemonset.updateStrategy | indent 4 }}
+    type: RollingUpdate
```
achandras:
```suggestion
    type: {{ .Values.daemonset.updateStrategy.type }}
```

charts/sysdig/templates/daemonset.yaml
lilx1ao: noted

charts/sysdig/templates/configmap.yaml
lilx1ao: good call, thanks!

charts/sysdig/templates/daemonset.yaml
xiao-sysdig: shall I do a null check?

charts/sysdig/templates/_helpers.tpl
```diff
@@ -231,3 +231,20 @@ Sysdig Eve Connector Secret generation (if not exists)
...
+{{- if .Values.daemonset.updateStrategy.rollingUpdate -}}
+    {{- if .Values.daemonset.updateStrategy.rollingUpdate.maxUnavailable -}}
+    {{- .Values.daemonset.updateStrategy.rollingUpdate.maxUnavailable -}}
+    {{- end -}}
+{{- else if eq .Values.resourceProfile "small" -}}
+    {{- 1 -}}
+{{- else if eq .Values.resourceProfile "medium" -}}
+    {{- 10 -}}
+{{- else if eq .Values.resourceProfile "large" -}}
+    {{- 20 -}}
```
achandras: can we have a fallback case here that's just the default `1`

lilx1ao: sure thing

charts/sysdig/templates/_helpers.tpl
```diff
@@ -231,3 +231,20 @@ Sysdig Eve Connector Secret generation (if not exists)
...
+*/}}
+{{- define "sysdig.parallelStarts" -}}
+{{- if .Values.daemonset.updateStrategy.rollingUpdate.maxUnavailable -}}
+    {{- .Values.daemonset.updateStrategy.rollingUpdate.maxUnavailable -}}
+{{- else if eq .Values.resourceProfile "small" -}}
+    {{- 1 -}}
+{{- else if eq .Values.resourceProfile "medium" -}}
+    {{- 10 -}}
+{{- else if eq .Values.resourceProfile "large" -}}
+    {{- 20 -}}
```
irozzo-1A: As we discussed offline, setting the `max_parallel_cold_starts` to 20 may add too much load on the Apiserver for big clusters.
Cold-start is implemented using k8s leader election mechanism based on leases resources. Each agent tries to acquire each lease every 15s during the bootstrap, this means that in a large cluster, say 1K nodes we will have ~1333rps in the worst case (1000 nodes * 20 leases / 15s). I would keep 10 for the moment that is more conservative, until we better assess the impact.

lilx1ao: thank you. I will change this value to 10 for all cluster with more than 10 nodes.

xiao-sysdig: @irozzo-1A did you suggest if someone from QA to test this?

achandras (approved):
lgtm

PR #379, opened 2022-03-30
[sysdig] update benchmarkRunner version and fix SA creation

achandras:

Prevents creation of serviceAccount, clusterrole, and clusterrolebinding when the node analyzer deployment is disabled.

## Checklist
- [x] Chart Version bumped


noahkraemer (changes requested):
The `deploy-nia` value is set based on the `nodeImageAnalyzer.deploy`, which refers to the legacy Node Image Analzer daemonset, which does not use the new SA & associated objects: https://github.com/sysdiglabs/charts/blob/master/charts/sysdig/templates/_helpers.tpl#L211-L213

I think we need a new `deploy-na` helper that is based on `nodeAnalyzer.deploy` instead

achandras:
Oh yep, missed that. I see the daemonset is using `and (not (include "deploy-nia" .)) .values.nodeanalyzer.deploy`, will make a helper using that

achandras:
Updated, thanks for the feedback

noahkraemer (approved):
Great, thanks for catching this @achandras!

aq17 (approved):

PR #390, opened 2022-04-08
[agent] auto endpoint for region

achandras:

Adds a lookup for collector endpoints if region is provided.

Also changes the lookup of values in `sysdig.settings` so that it simply returns a value only if the key isn't present. 

## Checklist
- [x] Chart Version bumped


charts/agent/templates/_helpers.tpl
```diff
@@ -137,18 +137,16 @@ Daemonset labels
...
+Use like: {{ include "get_if_not_in_settings" (dict "root" . "default" "<coalesced value>" "setting" "<agent_setting>") }}
+Return the default only if the value is not defined in sysdig.settings.<agent_setting>.
 */}}
-{{- define "get_or_fail_if_in_settings" -}}
-{{- $keyValue := tpl (printf "{{- .Values.%s -}}" .key) .root }}
-{{- if $keyValue -}}
-    {{- if hasKey .root.Values.sysdig.settings .setting }}{{ fail (printf "Value '%s' is also set via .sysdig.settings.%s'." .key .setting) }}{{- end -}}
-    {{- $keyValue -}}
+{{- define "get_if_not_in_settings" -}}
+{{- if not (hasKey .root.Values.sysdig.settings .setting) -}}
```
achandras, line 144: I found this to be easier to use when coalescing multiple values, but got rid of the fail condition. Can think of a different approach if that kind of message is something we want to keep.

charts/agent/README.md
```diff
@@ -63,57 +64,60 @@ The command removes all the Kubernetes components associated with the chart and
...
+| `sysdig.accessKey`                      | Your Sysdig Agent Access Key. Overrides `global.sysdig.accessKey`                                                                          | ` ` Either accessKey or existingAccessKeySecret is required |
+| `sysdig.existingAccessKeySecret`        | Alternatively, specify the name of a Kubernetes secret containing an 'access-key' entry. Overrides `global.sysdig.existingAccessKeySecret` | ` ` Either accessKey or existingAccessKeySecret is required |
+| `sysdig.disableCaptures`                | Disable capture functionality (see https://docs.sysdig.com/en/disable-captures.html)                                                       | `false`                                                     |
+| `sysdig.settings`                       | Additional settings, directly included in the agent's configuration file `dragent.yaml`. Overrides other settings.                         | `{}`                                                        |
```
achandras: Along the lines of the helper change, since I took out the message, I just noted here that anything specified with `sysdig.settings` will override other values.

airadier (approved):
Nice! This logic only applies to the collector setting in the agent, right? I think we will also need a similar (but not equal) value for the API endpoint for benchmarks, NIA, etc. so should we include this in the parent chart maybe, instead of in the agent? Or do we replicate the logic in the node-analyzer subchart?

See the example I was using when deploying agent via terraform for demo environments, which considers both agent and NA. 
https://github.com/sysdiglabs/demoenv-scenarios/blob/d9aedcc60eb50c4e0244f0ebb92878a286cca477/terraform/modules/agent/main.tf#L2
This TF code will be removed when the new helm chart is ready.

achandras:
This was actually inspired by your example! Since the secure API endpoint for NA is different, I think it makes sense to have a separate helper for it.

Also I wasn't able to find a good solution for inheriting common helpers from the parent chart and had to use a library chart instead, so the dependencies looked like `helper-chart`->`component-chart`->`parent-chart`. And unfortunately it seems like the current packaging workflow doesn't resolve multi-level dependencies, so that would have to be addressed for library charts. Do you have any experience/suggestions around this?

PR #392, opened 2022-04-08
[agent] use global namespace value if provided

achandras:

Provides options for installing namespace-scoped components into a different namespace.

## Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -5,7 +5,7 @@ description: Sysdig Monitor and Secure agent
 type: application
 
 # currently matching sysdig 1.12.70
-version: 1.2.0
+version: 1.3.0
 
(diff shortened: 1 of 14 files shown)


lilx1ao (approved):

charts/agent/Chart.yaml
```diff
@@ -5,7 +5,7 @@ description: Sysdig Monitor and Secure agent
 type: application
 
 # currently matching sysdig 1.12.56
-version: 1.0.2
+version: 1.0.3
```
lilx1ao: to 1.1.1 maybe?

achandras: thanks for pointing this out! I'll use 1.3.0 in line with the earlier versioning discussion

PR #413, opened 2022-05-04
[agent] remove the appchecks

lilx1ao:
## Checklist
- [x] Chart Version bumped


charts/agent/scripts/appchecks2helm
```diff
@@ -1,11 +0,0 @@
-#!/bin/bash
```
lilx1ao, line 1: there is nothing else in that /scripts directory other than the removed appchecks2helm file. shall I remove the directory altogether? I assume there is no code scanning through the subdirectory structures and so it is safe to do.

yashgiri (approved):
LGTM

lilx1ao:
tested with helm template and the Jenkins cluster. nothing seems to be broken from this

charts/agent/scripts/appchecks2helm
achandras: git doesn't keep empty directories, so since this was the only file under this folder it counts as deleting the folder too

achandras (approved):

charts/agent/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@
 
 This file documents all notable changes to the Sysdig Agent Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+## v1.1.1
```
achandras:
```suggestion
## v1.4.3
```

charts/sysdig/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: sysdig
-version: 1.13.3
```
achandras: this chart bump is not needed

achandras (changes requested):
Couple of minor changes needed, but otherwise looks good

charts/agent/scripts/appchecks2helm
lilx1ao: thanks! this is great to know.

charts/sysdig/Chart.yaml
lilx1ao: thanks

achandras (approved):

PR #415, opened 2022-05-10
[sysdig] Add CSPM templates

chen-shmilovich-sysdig:
This PR introduces two new components to the `sysdig` chart:
* CSPM Analyzer - A container inside the existing `NodeAnalyzer` Daemonset. It analyzes and collects K8s node configurations.
*  CSPM Collector - Deployment to collect K8s resource manifests.

## Checklist
- [X] Chart Version bumped


charts/sysdig/templates/clusterrole.yaml
```diff
@@ -105,4 +105,23 @@ rules:
     verbs:
       - "use"
 {{- end }}
+{{- if .Values.cspmCollector.deploy  }}
```
achandras: This might make sense to put as a separate clusterrole + clusterrolebinding to the service account, since this is only for the cspm collector, not the agent

charts/sysdig/templates/cspmCollector/deployment-cspm-collector.yaml
```diff
@@ -0,0 +1,85 @@
...
+      app.kubernetes.io/name: {{ include "sysdig.name" . }}-cspm-collector
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ include "sysdig.name" . }}-cspm-collector
+        app.kubernetes.io/instance: {{ .Release.Name }}
+{{ include "sysdig.labels" . | indent 8 }}
+    spec:
+      serviceAccountName: {{ template "sysdig.serviceAccountName" .}}
```
achandras: following on from the other comment, I think it may make sense to have a separate service account for this deployment. This will make it easier to split out later when moving these changes to the node-analyzer chart

charts/sysdig/templates/clusterrole.yaml
chen-shmilovich-sysdig: Thanks! I added a clusterole and a clusterrolebinding for the cspm collector.

charts/sysdig/templates/cspmCollector/deployment-cspm-collector.yaml
chen-shmilovich-sysdig: Agree. I changed it so that the cspm-collector has its own service account.

achandras (approved):
lgtm!

PR #419, opened 2022-05-12
[cspm-collector] CSPM Collector chart

chen-shmilovich-sysdig:
## Checklist
- [X] Chart Version bumped


chen-shmilovich-sysdig:
@yashgiri @achandras Last one before I can move on and update "sysdig-deploy". Thank you and sorry for the hassle.

chen-shmilovich-sysdig:
@yashgiri @achandras Last one before I can move on and update "sysdig-deploy". Thank you and sorry for the hassle.

charts/cspm-collector/templates/clusterrole.yaml
```diff
@@ -0,0 +1,56 @@
+{{- if and .Values.rbac.create .Values.deploy  }}
```
achandras: the `deploy` flag is probably unnecessary, since a user could just as easily not deploy the chart 😄 

this can be handled by a condition in the parent chart, as we do for the current subcharts

achandras (approved):
one minor comment, but otherwise lgtm

charts/cspm-collector/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+```bash
+$ helm install --namespace cspmcollector cspm-collector -f values.yaml sysdig/cspm-collector
+```
+
+Alternatively, a YAML file that specifies the values for the parameters can be provided while installing the chart. For example,
+
+```bash
+$ helm install my-release-name -f values.yaml sysdig/cspm-collector
+```
```
yashgiri, line 66: Maybe could also include steps for helm upgrade.

charts/cspm-collector/templates/clusterrole.yaml
chen-shmilovich-sysdig: removed :)

charts/cspm-collector/README.md
chen-shmilovich-sysdig: @yashgiri wouldn't it be the same, just "upgrade" instead of "install"?
I don't see that we include it in other charts.🤔

yashgiri (commented):
LGTM

charts/cspm-collector/README.md
yashgiri: You are right about the fact that it's the same and also it's not included in most of the other charts. I had to modify it sometime back as upgrade documentation was lacking for 2 major charts and hence my 2c. I would refer to @achandras on this.

chen-shmilovich-sysdig: I see, thanks for bringing this to my attention.
We plan to refactor the readme file and add a link to the official docs once they'll get released.
I'll add there the upgrade instructions as well.

PR #420, opened 2022-05-13
[sysdig] Added ability to selectively deploy benchmark runner and host analyzer

kadkins-sysdig:

Added ability to selectively deploy benchmark runner and host analyzer

## Checklist
- [X] Chart Version bumped


airadier (commented):
My 2 cents: I would consider the case when the value is not defined, as customers upgrading from a previous version, with the --reuse-values behavior (default if you perform a plain "helm upgrade") might have issues if the value is not defined.

kadkins-sysdig:
As suggested by @airadier, I set defaults to true in _helpers.tpl.

charts/sysdig/templates/_helpers.tpl
```diff
@@ -219,6 +219,12 @@ true
 {{- end -}}
 {{- end -}}
 
+{{/*
+Set default to true for deployment of the benchmarkRunner and the hostAnalyzer
+*/}}
+{{ default true .Values.nodeAnalyzer.benchmarkRunner.deploy }}
```
achandras, line 225: I am not sure that this will have the desired result, as the value is not being set/used. But also one of the quirks of the `default` function is that `undefined`/`nil` and `false` are treated the same (ie the statement above will always evaluate true, even when the value is defined), which makes using it to check for undefined keys a bit tricky. This really rears its head is when we want an undefined bool to default to `true`, such as when handling `--reuse-values` like @airadier said.

I think the best approach here would be to define a helper which checks if the `.Values.nodeAnalyzer.benchmarkRunner.deploy` key exists, and return `true` if not, and use that helper in place of checking the deploy flag in all the templates. Same for the `.Values.nodeAnalyzer.hostAnalyzer.deploy` flag below.

PR #427, opened 2022-05-23
[sysdig] add customizable envvars for NodeAnalyzer deployments

r2bit:
Recently, while working on a performance issue with Sysdig support, we
discovered that tuning RuntimeScanner's memory settings fixed the issue.
However, that tunable is set via environment variables and current Chart
does not allow setting custom envvars.
Instead of having to wait until that specific tunable is added to the
Chart, I think it's better to add a broad support for custom envvars for
just such occasions.

This PR adds `env` object to:
* imageAnalyzer
* hostAnalyzer
* benchmarkRunner
* runtimeScanner
* cspmAnalyzer
* cspmCollector

## Checklist
- [x] Chart Version bumped


yashgiri (commented):
LGTM

charts/sysdig/values.yaml
```diff
@@ -553,6 +565,8 @@ cspmCollector:
       cpu: 500m
       memory: 1536Mi
 
+  env: { }
```
lilx1ao: we do not leave spaces in between the curly brackets as the default value placeholder. check out the example below from line 556/570.

achandras:
@r2bit thanks for this PR! Could you please bump the version in `Chart.yaml`?

r2bit:
> @r2bit thanks for this PR! Could you please bump the version in `Chart.yaml`?

Hope patch bump is enough. :)

achandras:
Thanks!

achandras (approved):

PR #451, opened 2022-06-09
[node-analyzer] Rename CSPM to KSPM

chen-shmilovich-sysdig:
## What this PR does / why we need it:
Rename CSPM to KSPM

## Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [X] Chart Version bumped
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[mychartname]`)
- [X] PR only contains changes for one chart

PR #457, opened 2022-06-10
[node-analyzer] update node analyzer chart 1.14.34

achandras:

Update chart to match sysdig-1.14.34. I believe #451 should bring this the rest of the way up to date.

## Checklist
- [x] Chart Version bumped


charts/node-analyzer/values.yaml
```diff
@@ -92,6 +92,9 @@ nodeAnalyzer:
     - effect: NoSchedule
       key: node-role.kubernetes.io/master
 
+  # Set nodeAnalyzer daemonset priorityClassName
+  priorityClassName:
```
lilx1ao, line 96: no {} here?

achandras: since this is supposed to be a string, we could set a default `""`, but it's not strictly necessary since it only gets templated if the value is [set](https://github.com/sysdiglabs/charts/blob/master/charts/node-analyzer/templates/daemonset-node-analyzer.yaml#L66)

lilx1ao: ah

lilx1ao (approved):

charts/node-analyzer/README.md
```diff
@@ -180,7 +185,18 @@ $ helm install --namespace nodeanalyzer node-analyzer -f values.yaml sysdig/node
...
+
+The Node Analyzer is deployed by default unless you set the value `nodeAnalyzer.deploy` to `false`.
+
+The Node Analyzer daemonset contains three containers, each providing a specific functionality. This daemonset replaces
+the (deprecated) Node Image Analyzer daemonset.
+
+See
+the [Node Analyzer installation documentation](https://docs.sysdig.com/en/node-analyzer--multi-feature-installation.html)
+for details about installation, and
+[Running Node Analyzer Behind a Proxy](https://docs.sysdig.com/en/node-analyzer--multi-feature-installation.html#UUID-35c14c46-b327-c2a8-ed9c-82a2af995218_section-idm51621039128136)
```
yashgiri: Not really an issue but both the links point to the same page starting at the top although both seem to have different links.

yashgiri (approved):

charts/node-analyzer/README.md
achandras: yup, these look outdated. fixed this and a few other instances of outdated anchors. thanks!

PR #467, opened 2022-06-16
[cloud-connector] allow role based access

sledigabel:
The only current way to connect to the ingestors in AWS are through
a User and generate an access key and a secret.

However, kubernetes workloads can also assume a role at startup time (for example, [kube2iam](https://github.com/jtblin/kube2iam)) without the need to impersonate a user. IAM Users are limited and their secrets are difficult to rotate. On the other hand, assuming an IAM role generates temporary credentials at each assume and is safe to use.

In the current implementation, the secret will always include the AWS credentials (even if empty) in the secret, and the deployment always loads it in an environment variable at load time.

That causes a problem as in the AWS SDK, the env vars will take precedence over the [role assumption](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials.html).
This PR introduces a new toggled called `aws.roleBasedAccess`, it doesn't add the accessKeyId and secretAccessKey into the secret and it doesn't mount it from the deployment, leaving the connector to assume the IAM role.

The current setup doesn't allow IAM role based access because it always sets the AWS_... environment variables.
This PR allows those variables not to be set if a flag is enabled.

## Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/cloud-connector/Chart.yaml
+++ b/charts/cloud-connector/Chart.yaml
@@ -3,7 +3,7 @@ name: cloud-connector
 description: Sysdig Cloud Connector
 
 type: application
-version: 0.7.8
+version: 0.7.9
 appVersion: 0.16.10
 home: https://sysdiglabs.github.io/cloud-connector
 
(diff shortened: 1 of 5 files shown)


sledigabel:
Upon further testing, it looks like if the var is empty the SDK is smart enough to ignore it. I'll close that PR for now :-) sorry for the noise

PR #477, opened 2022-06-27
[sysdig/sysdig-deploy] helm chart notes modified with region based links

yashgiri:
## Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/sysdig-deploy/templates/_helpers.tpl
+++ b/charts/sysdig-deploy/templates/_helpers.tpl
@@ -1,0 +1,37 @@
+{{/*
+Determine sysdig monitor endpoint based on provided region
+*/}}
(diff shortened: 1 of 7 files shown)


charts/sysdig-deploy/templates/_helpers.tpl
```diff
@@ -0,0 +1,37 @@
+{{/*
+Determine sysdig monitor endpoint based on provided region
+*/}}
+{{- define "global.monitorUrl" -}}
```
achandras: I think the `global` here and in the helper below is misleading, just `monitorUrl` is probably fine

achandras (approved):
lgtm other than a minor comment

achandras (approved):

PR #480, opened 2022-06-29
[agent] use nil check instead of emptiness

achandras:

Previously, the emptiness check would prevent boolean "false" values from being printed

## Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/agent/templates/_helpers.tpl
+++ b/charts/agent/templates/_helpers.tpl
@@ -143,7 +143,7 @@ Return the default only if the value is not defined in sysdig.settings.<agent_se
 */}}
 {{- define "get_if_not_in_settings" -}}
 {{- if not (hasKey .root.Values.sysdig.settings .setting) -}}
-  {{- if ne .default nil -}}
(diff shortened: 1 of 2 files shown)


charts/agent/templates/_helpers.tpl
```diff
@@ -143,7 +143,7 @@ Return the default only if the value is not defined in sysdig.settings.<agent_se
 */}}
 {{- define "get_if_not_in_settings" -}}
 {{- if not (hasKey .root.Values.sysdig.settings .setting) -}}
-  {{- if .default -}}
+  {{- if and (ne .default nil) (printf "%s" .default) -}}
```
lilx1ao: will things be printed here??

lilx1ao (approved):

charts/agent/templates/_helpers.tpl
achandras: it's more like sprintf, it'll all be contained in that temporary expression. but the old helm 3.4 doesn't like this approach anyway so going to try something different

PR #484, opened 2022-07-04
[admission-controller] update description on scanner.enabled

andrewd-sysdig:
To explain that disabling this disables the whole admission controller itself (and just leaves the Kubernetes Audit functionality.

## Checklist
- [ ] Chart Version bumped
- [ ] Title of the PR starts with chart name (e.g. `[mychartname]`)


ElenaFerante (approved):

wideawakening:
maybe its a good moment to evaluate the default values of

- `scanner.enabled` default to true
  -  IMHO, better set to false, as user needs to enable it on SDS for it to work. if they don't do it, deployed resources make no sense
-  `features.k8sAuditDetections` default to false
   - IMHO, better set to true, as is the mail feature that would work out of the box.
   
are there any other "main features" of the AC we should dig into?

andrewd-sysdig:
Good call - that makes sense to me, and no other main features as far as i am aware

tembleking:
Replaced by #499

PR #542, opened 2022-07-27
add /etc -> /host/etc volume bind

deepskyblue86:
After integrating falcosecurity/libs/pull/220, the agent can have an up-to-date users and groups list from `/etc/passwd` and `/etc/group`.
To have those files in the container we mount `/etc` in `/host/etc` and symlink them in the container `/etc`.
If we mount them individually, they will become stale in case of an inode change, which is what happens with `useradd`.

## Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/agent/CHANGELOG.md
+++ b/charts/agent/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 This file documents all notable changes to the Sysdig Agent Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+## v1.5.10
+
+### Minor changes
+* Added /etc to container /host/etc volume bind
+
 ## v1.5.9
 
 ### Bugfix
--- a/charts/agent/templates/daemonset.yaml
+++ b/charts/agent/templates/daemonset.yaml
@@ -117,9 +117,6 @@ spec:
               name: sys-tracing
               readOnly: true
             {{- end }}
-            - mountPath: /host/etc/os-release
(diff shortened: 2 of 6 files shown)


charts/agent/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 This file documents all notable changes to the Sysdig Agent Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+### Minor changes
```
achandras, line 9: Please add version heading above

charts/agent/templates/daemonset.yaml
```diff
@@ -175,6 +175,9 @@ spec:
               name: modprobe-d
               readOnly: true
             {{- end }}
+            - mountPath: /host/etc
```
achandras, line 175: with this addition I think the `osrel` and `etc-fs` mounts/volumes can be removed

charts/sysdig/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 This file documents all notable changes to Sysdig Helm Chart. The release numbering uses [semantic versioning](http://semver.org).
 
+### Minor changes
```
achandras, line 8: version heading

achandras:
Each chart's `Chart.yaml` will need a minor version bump

lilx1ao (approved):

PR #545, opened 2022-07-27
[registry-scanner] Fix templating error (#545)

firaxis:
Hi!
This fixes templating error.
I try to install this chart with flux, but I get following error
```
  Warning  error   7s (x6 over 31s)  helm-controller  Helm install failed: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors:
  line 29: mapping key "app.kubernetes.io/name" already defined at line 27
  line 30: mapping key "app.kubernetes.io/instance" already defined at line 28
```
Looks like selectorLabels that is valid for deployments in .spec.selector.matchLabels, accidentally got into a cronjob.

## Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/registry-scanner/templates/cronjob.yaml
+++ b/charts/registry-scanner/templates/cronjob.yaml
@@ -23,10 +23,7 @@ spec:
         metadata:
           name: {{ include "registry-scanner.fullname" . }}
           labels:
-            app.kubernetes.io/name: {{ include "registry-scanner.name" . }}
-            app.kubernetes.io/instance: {{ .Release.Name }}
-            {{- include "registry-scanner.selectorLabels" . | nindent 12 }}
-            {{- include "registry-scanner.customLabels" . | nindent 12 }}
+            {{- include "registry-scanner.labels" . | nindent 12 }}
           {{- with .Values.podAnnotations }}
           annotations:
             {{- toYaml . | nindent 12 }}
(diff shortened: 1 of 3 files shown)


airadier:
Thanks very much! Let me double check it, because apparently those labels have been there for months, causing a duplicate `app.kubernetes.io/name`but it was working... I'll run some tests and try to find what's wrong.

charts/registry-scanner/templates/cronjob.yaml
```diff
@@ -25,7 +25,6 @@ spec:
           labels:
             app.kubernetes.io/name: {{ include "registry-scanner.name" . }}
             app.kubernetes.io/instance: {{ .Release.Name }}
-            {{- include "registry-scanner.selectorLabels" . | nindent 12 }}
```
airadier, line 28: I see the `selectorLabels`already defines `app.kubernetes.io/name` and `app.kubernetes.io/instance` labels, and seems to be the standard way in _helpers.tpl and other Helm charts to define the standard set of labels. It is used in the deployment template, and then in the service to match or "select" the corresponding pods by labels.

What I would suggest is keeping the include (so the line you just removed) and instead remove lines 26 and 27, as those labels are already provided in the `registry-scanner.selectorLabels` template, what do you think?

airadier (changes requested):

charts/registry-scanner/templates/cronjob.yaml
firaxis: Thanks for suggestion,
I've found `registry-scanner.labels` that includes `registry-scanner.selectorLabels` this should be the way.

firaxis:
Closing in favor https://github.com/sysdiglabs/charts/pull/557

PR #326, opened 2022-01-12
Update daemonset.yaml to add the https_proxy and http_proxy in the env section of the sysdig-agent-kmodule container

luiof:
## What this PR does / why we need it:

Added the missing https_proxy and http_proxy in the env section of the sysdig-agent-kmodule container.
We need it since it is missing in the chart but it is supported by the container (we tested it).

## Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [ ] Chart Version bumped
- [ ] Variables are documented in the README.md
- [ ] Title of the PR starts with chart name (e.g. `[mychartname]`)
- [ ] PR only contains changes for one chart

PR #572, opened 2022-08-09
Added proxy support for agent initContainer

dark-vex:
Signed-off-by: Daniele De Lorenzi <daniele.delorenzi@sysdig.com>

This PR introduce the proxy support for agent initContainer, right know the proxy is added only for agent container but not for agent-kmodule so when a customer is using a proxy and agent-slim (which is default) the kernel module cannot be downloaded.

There is also an old PR with the same modification here:
https://github.com/sysdiglabs/charts/pull/326
but it's quite old so I did it from scratch

## Checklist

- [X] Chart Version bumped
- [X]  Check GithubAction checks (like lint) to avoid merge-check stoppers


achandras (approved):
lgtm, thanks for doing this!

yashgiri (approved):

PR #574, opened 2022-08-09
ci: Mark inactive PRs and Issues as stale and close them automatically

tembleking:

Marks the PRs and Issues as Stale if they have been inactive for more than 60 days. It closes them after 7 days after being marked as stale.
This will ensure we don't have inactive PRs in this repository that could just add noise.


wideawakening (approved):
seems legit.
we need some chart ownership

PR #582, opened 2022-08-21
[sysdig-deploy] Remove default for global.imageRegistry

achandras:

Introducing a default value for `global.imageRegistry` makes chart-specific image registry settings not work. Removing the default goes back to previous behavior.

## Checklist

- [x] Chart Version bumped


lilx1ao (approved):

lilx1ao:
is it because the value can also be in docker?

achandras:
Sort of, but mainly because the logic for this is a little weird, since we carried it over from the old chart... if the field is defined at all on globals, it takes precedence over any local value.

Probably want to rethink this at some point.

PR #583, opened 2022-08-22
[agent] move ingresses clusterrole's resource in networking.k8s.io group

fremmi:

The clusterrole's ingresses resources has been moved from extensions to networking.k8s.io
There is no need to maintain the extensions group as it is not used not even in older agent versions

## Checklist

- [x] Chart Version bumped


achandras (approved):

achandras (approved):
lgtm!

PR #595, opened 2022-08-29
[nodeanalyzer]: add host scanner component

mech-pig:
## Checklist

- [ ] Chart Version bumped


charts/node-analyzer/README.md
```diff
@@ -139,6 +140,19 @@ The following table lists the configurable parameters of the Sysdig Node Analyze
 | `nodeAnalyzer.benchmarkRunner.resources.limits.cpu`                  | Benchmark Runner CPU limit per node                                                      | `500m`                                                                         |
 | `nodeAnalyzer.benchmarkRunner.resources.limits.memory`               | Benchmark Runner Memory limit per node                                                   | `256Mi`                                                                        |
 | `nodeAnalyzer.benchmarkRunner.env`                                   | Extra environment variables that will be passed onto pods                                | `{}`                                                                           |
+| `nodeAnalyzer.hostScanner.deploy`                                 | Deploy the Host Scanner                                                               | `false`                                                                        |
```
mnovait: shouldn't it be set to true by default?

mnovait: I answer it by myself: no because this is the internal chart.

charts/node-analyzer/templates/daemonset-node-analyzer.yaml
```diff
@@ -629,13 +637,121 @@ spec:
...
+      {{- end }}
       {{- with .Values.nodeAnalyzer.nodeSelector }}
       nodeSelector: {{ toYaml . | nindent 8 }}
       {{- end }}
       {{- with .Values.nodeAnalyzer.affinity }}
       affinity: {{ toYaml . | nindent 8 }}
       {{- end }}
 {{- end }}
-{{- end }}
+{{- end }}
```
dippie8, line 757:
```suggestion
{{- end }}

```

dippie8 (approved):
left a minor comment, approved meanwhile

charts/node-analyzer/templates/_helpers.tpl
```diff
@@ -222,8 +229,15 @@ nodeAnalyzer agentConfigmapName
     {{- default .Values.global.agentConfigmapName | default "sysdig-agent" -}}
 {{- end -}}
 
-{{- define "nodeAnalyzer.deployHostAnalyzer" -}}
```
yashgiri, line 225: I assume nodeAnalyzer.deployHostAnalyzer is being removed and changed for nodeAnalyzer.deployHostscanner if that is the case maybe we can remove nodeAnalyzer.hostAnalyzer.deploy from the readme.
Edit: change not needed as it is present below.

yashgiri (approved):
Hey, Thank you for the changes and for adding a test file as well. I see there are changes for nodeAnalyzer.deployHostScanner that I believe can be added to the test file conditional_flag_test.yaml. Let me know what do you think?  Otherwise it is lgtm

charts/node-analyzer/templates/_helpers.tpl
ProvoK: HostAnalyzer and HostScanner are two different components, so both shall remain :)

ProvoK:
@yashgiri If it doesn't bother you I'd leave the tests in a separate file just because are a bit easier to understand/look for, even if they're testing "something similar"

yashgiri:
That is fine by me. LGTM.

charts/node-analyzer/templates/daemonset-node-analyzer.yaml
guidobonomi, line 757:
```suggestion
{{- end }}

```

charts/node-analyzer/templates/daemonset-node-analyzer.yaml
```diff
@@ -213,9 +216,10 @@ spec:
             readOnly: true
           - name: tmp-vol
             mountPath: /host/tmp
-{{- end }}
-{{ if not .Values.secure.vulnerabilityManagement.newEngineOnly }}
```
guidobonomi, line 217: @ProvoK it's complex to me to properly understand this change from the PR but can you please just double check that the removal of this `if` is correct?

ProvoK: it is, because  `newEngineOnly` is "included" in the new helper "nodeAnalyzer.deploImageAnalyzer" defined here:https://github.com/sysdiglabs/charts/pull/595/files#diff-c0c8ff035f55b2bed4c0503bbcb2d06cd9cc71064dee11c6ccd412dc0dcb3fafR251

PR #598, opened 2022-08-29
[agent] Add kubectl rollout test with 5m timeout

achandras:

This adds a rudimentary test for the chart testing suite - `kubectl rollout status` with a timeout of 5 minutes.

Unfortunately, `kind` was not suitable for running the agent, so I switched the testing cluster to `k3s`. There should be no difference in actual use, and setup times for both are fairly similar.

## Checklist

- [x] Chart Version bumped


yashgiri (approved):
Thank you for this change.

lilx1ao:
great work! thank you

lilx1ao (approved):

PR #603, opened 2022-09-01
Update commands to used existing variables

0snug0:
Made all the variables bash variables

SDC_ACCESS_KEY has generally been used in many other places, so this is why that one specifically has SDC added

## Checklist

- [ ] Chart Version bumped


achandras:
@0snug0 thanks for the updates and the context around the naming. The same variables are used in the `Installation` section as well, it probably makes sense to update those as well, right?

0snug0:
Yes agreed, those should also be updated

PR #604, opened 2022-09-02
[cloud-connector] feat: Add feature-toggle for bruteforce detection

penguinjournals:
## Checklist

- [x] Chart Version bumped


charts/cloud-connector/values.yaml
```diff
@@ -127,6 +127,8 @@ ingestors: []
 scanners: []
 
 bruteForceDetection:
+  # Enable Brute Force detection
```
wideawakening, line 130: since this is the only available documentation right now for the user, could we briefly explain what it does?

PR #631, opened 2022-09-16
[agent][sysdig-deploy] Make the agent use tags from global.sysdig.tag…

yashgiri:
…s when available

## Checklist

### DIFF
--- a/charts/agent/templates/_helpers.tpl
+++ b/charts/agent/templates/_helpers.tpl
@@ -252,3 +252,18 @@ Deploy on GKE autopilot
         true
     {{- end -}}
 {{- end -}}
+
+{{/*
+Use global sysdig tags for agent
+*/}}
+{{- define "agent.tags" -}}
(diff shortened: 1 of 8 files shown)


charts/agent/templates/_helpers.tpl
```diff
@@ -242,4 +242,20 @@ Agent agentConfigmapName
 */}}
 {{- define "agent.configmapName" -}}
     {{- default .Values.global.agentConfigmapName | default "sysdig-agent" -}}
+{{- end -}}
+
+{{/*
+Use global sysdig tags for agent
+*/}}
+{{- define "sysdig.settings.tags" -}}
```
achandras: Nit: This is probably fine, but given that we also have a value with this name, maybe we can call this something simple like `agent.tags`

yashgiri:
LGTM but can't approve. will probably need @lilx1ao 's approval.

lilx1ao (approved):

PR #653, opened 2022-10-13
[harbor-scanner-sysdig-secure] Introduced customEntryPoint

dark-vex:
Signed-off-by: Daniele De Lorenzi <daniele.delorenzi@sysdig.com>

This PR introduce the possibility to override the container entryPoint via `values.yaml file`, it's useful when you're using for example HashiCorp Vault for inject the secrets, ex:
```
customEntryPoint: ["/bin/sh", "-c", "source /vault/secrets/secret-file && /<path-to-entrypoint.sh>"]
```
I have created also the changelog file which was missing for this chart

## Checklist

- [X] Title of the PR starts with chart name (e.g. [mychartname])

### DIFF
--- a/charts/harbor-scanner-sysdig-secure/CHANGELOG.md
+++ b/charts/harbor-scanner-sysdig-secure/CHANGELOG.md
@@ -1,0 +1,21 @@
+# Chart: Harbor Scanner for Sysdig Secure
+
+## Change Log
+
+## v0.3.3
+
+### Feature
+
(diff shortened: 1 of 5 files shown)


airadier (approved):
LGTM. But is using a custom entrypoint the recommended way to inject Vault secrets, or should it be done via a sidecar?

dark-vex:
@airadier there are different methods with Vault but we have a customer that cannot use k8s secrets and can only use the entrypoint override method

PR #663, opened 2022-10-18
[admission-controller] Add option to not create rbac roles

anderssonw:

Currently the admission controller does not allow you do not create rbac roles for the webhook. Our team wants to create these clusterroles ourselves to not over-privilege the SA we have for sysdig, and therefore this is a useful addition.

Could be an idea to do this globally for the admission controller instead of separately for the webhook and scanner.

## Checklist


charts/admission-controller/values.yaml
```diff
@@ -119,6 +119,10 @@ webhook:
     # Configure port for the webhook service
     port: 443
 
+  rbac:
+    # Whether to create a clusterrole and bind it
```
legobrick: This and README.md description in table need to be the same. Mind changing this to `Enable the creation of ClusterRoles and the binding of these roles` ?

legobrick (changes requested):

charts/admission-controller/values.yaml
anderssonw: Done :)

anderssonw:
Not entirely sure why the pre-commit fails. I attempted to run it locally and the linting is fine, but not sure how to run the docs.

dark-vex:
Thanks @anderssonw for your contribution! We added a couple of more changes to this other PR #664 we have cherry-picked your commits, so you're still the initial contributor :)

PR #684, opened 2022-11-02
[Sysdig] enable ksmp deploy by default

sameer-in:
## enable ksmp deploy by default


claffin:
If we change this to enabled by default, then next time a customer upgrades the Helm chart they're going to end up installing KSPM whether they actually wanted to or not. I don't think we want to do this? 

We should instead update the agent Helm install on data sources page, in the readme and docs to include `--set kspm.deploy=true`.

aroberts87:
I agree with @claffin on this one. I don't think we want to change the defaults. The `sysdig-deploy` README shows how to enable kspm if anyone wants to deploy it. https://github.com/sysdiglabs/charts/tree/master/charts/sysdig-deploy#configuration

yashgiri:
+1 @claffin @aroberts87

sameer-in:
I agree @claffin. Let me close this PR.

PR #695, opened 2022-11-08
CI workflow that updates sysdig deploy chart version

yashgiri:

This PR includes the CI workflow changes that get activated when a new commit is pushed to the master branch. This workflow checks if any sysdig-deploy subcharts (agent, nodeanalyzer, kspm collector, rapid response) are updated. The workflow executes a script that checks for version changes in the sub-charts and depending on that increases the version of sysdig-deploy appropriately. 

## Checklist


scripts/sysdig-deploy/chart-version-bump.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
+
+set -e
+
+na_chart_path="charts/node-analyzer"
```
ironashram: is it really needed to shorten node-analyzer to na here? i would prefer to avoid it if possible

.github/workflows/update-sysdig-deploy.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Run the script that checks and modifies the sysdig-deploy version if dependencies are updated
+        run: scripts/sysdig-deploy/chart-version-bump.sh
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+      - name: Commit and push
+        run: | 
+            git add charts/sysdig-deploy/Chart.yaml
+            git commit -m "Updating sysdig-deploy chart version" && git push || true
```
ironashram: nit: for POSIX compliance we should always have a newline at the end of files

scripts/sysdig-deploy/chart-version-bump.sh
```diff
@@ -0,0 +1,93 @@
...
+    then
+        new_patch=$(($new_patch + 1))
+    fi
+
+    final_sysdig_deploy_version=$(echo "$new_major.$new_minor.$new_patch")
+    echo "Final deploy version is to be : $final_sysdig_deploy_version"
+    sed -i -e "1,/version: .*/ s/version: .*/version: ${final_sysdig_deploy_version}/" "$sysdig_deploy_path/Chart.yaml"
+}
+
+output=$(git diff --name-only HEAD~1 HEAD -- $na_chart_path $agent_chart_path)
```
ironashram: do i understand correctly that this will trigger the bump only if the version is changed in node-analyzer or agent? if yes is it correct?

scripts/sysdig-deploy/chart-version-bump.sh
```diff
@@ -0,0 +1,93 @@
...
+        echo "$chart"
+        changes=$(git diff HEAD~1 HEAD -- "$chart/Chart.yaml")
+        if ! [ -z "$changes" ]
+        then
+            check_upgrade_type $changes
+        fi
+    done    
+fi
+
+update_sysdig_deploy_chart
```
aroberts87: nit: as @ironashram mentioned above, a newline here would be nice.

.github/workflows/update-sysdig-deploy.yaml
```diff
@@ -0,0 +1,24 @@
+name: Update sysdig-deploy chart version if dependencies updated
+on:
+  push:
+    branches:
+      - master
+jobs:
+  update-sysdig-deploy:
```
aroberts87: nit: `update-sysdig-deploy-version`

.github/workflows/update-sysdig-deploy.yaml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout the repository
+        uses: actions/checkout@v2
+        with:
+        # fetch depth is needed since we will be taking git diff with HEAD^1
+          fetch-depth: 5
+      - name: Run the script that checks and modifies the sysdig-deploy version if dependencies are updated
+        run: scripts/sysdig-deploy/chart-version-bump.sh
```
aroberts87: Does the workflow short circuit if there is not version bump?

scripts/sysdig-deploy/chart-version-bump.sh
yashgiri: i will have to correct that and add the other paths too

yashgiri:
I had been able to test the changes on a private repository with na (nodeanalyzer) chart and agent chart. I have modified the script to extend it to the current dependency charts. 
The workflow gets executed on every push event to master. one of the job's steps further checks for file changes in any of the dependency charts. if changes are found the script to update sysdig-deploy chart.yaml gets executed. 
This script will make the appropriate version changes and the last PR job step will create a PR with the same changes.

scripts/sysdig-deploy/update-sysdig-deploy.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
+
+set -e
+
+node_analyzer_chart_path="node-analyzer"
+agent_chart_path="agent"
+kspm_collectore_chart_path="kspm-collector"
```
iurly: I guess you meant `kspm_collector_chart_path`?

scripts/sysdig-deploy/update-sysdig-deploy.sh
```diff
@@ -0,0 +1,88 @@
...
+
+    echo "subchart version in" $chart"/Chart.yaml"
+    echo $new_subchart_version
+    echo $chart "subchart version in sysdig-deploy/Chart.yaml"
+    echo $sysdig_subchart_version
+
+
+    IFS="."
+    read -ra prev_ver <<< "$sysdig_subchart_version"
+    read -ra next_ver <<< "$new_subchart_version"
```
iurly: I guess this works perfectly fine as is -- still, I'd try to avoid any subtle side effects by setting IFS locally:
```suggestion
    IFS="." read -ra prev_ver <<< "$sysdig_subchart_version"
    IFS="." read -ra next_ver <<< "$new_subchart_version"
```

scripts/sysdig-deploy/update-sysdig-deploy.sh
```diff
@@ -0,0 +1,88 @@
...
+minor=0
+major=0
+patch=0
+
+check_update_needed () {
+    
+    new_subchart_version=$(yq eval '.version' charts/$chart/Chart.yaml)
+    sysdig_subchart_version=$(yq eval '.dependencies[] | select(.name == "'$chart'") | .version ' charts/$sysdig_deploy_path/Chart.yaml | sed "s/~//")
+
+    echo "subchart version in" $chart"/Chart.yaml"
```
iurly: I believe this should do the trick (but please correct me if I'm wrong!)
```suggestion
    echo "subchart version in ${chart}/Chart.yaml"
```

scripts/sysdig-deploy/update-sysdig-deploy.sh
```diff
@@ -0,0 +1,88 @@
...
+                major=1
+            elif [ $idx -eq 1 ]
+            then
+                minor=1
+            elif [ $idx -eq 2 ]
+            then
+                patch=1
+            fi
+        fi
+    done
```
iurly, line 46: Feel free to ignore the following
```suggestion
    [ ${next_ver[0]} -gt ${prev_ver[0]} ] && major = 1
    [ ${next_ver[1]} -gt ${prev_ver[1]} ] && minor = 1
    [ ${next_ver[2]} -gt ${prev_ver[2]} ] && patch = 1
```

.github/workflows/update-sysdig-deploy-chart.yaml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          files: |
+            charts/node-analyzer/*
+            charts/agent/*
+            charts/kspm-collector/*
+            charts/rapid-response/*
+            charts/admission-controller/*
+      - name: Install YQ
+        if: steps.dependent_files.outputs.any_changed == 'true'
+        uses: dcarbone/install-yq-action@v1.0.1
```
iurly, line 27: I'm not particularly fond of those duplicates:
a) paths being duplicated on both the `.yaml` file AND the shell script
b) the gating condition `steps.dependent_files.outputs.any_changed == 'true'` all over the steps

For a) I would try to shift the responsibility onto the script so that it returns `0` when `No modifications necessary` and `1` in all other cases.
For b) it looks like an early-exit action would be the best solution, which is much-desired yet not available: https://github.com/actions/runner/issues/662.  A sub-optimal solution would be an environment variable as suggested here: https://github.com/actions/runner/issues/662#issuecomment-1295229779

iurly (approved):
Just a few suggestions -- otherwise looking good!

scripts/sysdig-deploy/update-sysdig-deploy.sh
yashgiri: I tried this change and got this error so ill probably skip this code change
```
yash@yash charts % ./scripts/sysdig-deploy/update-sysdig-deploy.sh 
subchart version in node-analyzer/Chart.yaml
1.7.33
node-analyzer subchart version in sysdig-deploy/Chart.yaml
1.7.32
patch: **** Can't open patch file 1 : No such file or directory
```

iurly: Oh I see, typical mistake.
There needs to be no space around the equal sign, i.e. patch=1

.github/workflows/update-sysdig-deploy-chart.yaml
yashgiri: Hello for a) I can remove the paths from the workflow the only thing I can foresee is that it would run this workflow on any change made on this repository as a push to master, which should be fine since it won't make any changes. Let me know if that is preferable.
Regarding return 0 which means execution is a success is fine but if I return 1 shouldn't that lead to erroring out and the workflow not running successfully?
b) I am assuming what is meant here is to refer to an environment variable after setting it once instead of the whole path. ill make that change now.

scripts/sysdig-deploy/update-sysdig-deploy.sh
yashgiri: I tried this change again after your correction
```
--- a/scripts/sysdig-deploy/update-sysdig-deploy.sh
+++ b/scripts/sysdig-deploy/update-sysdig-deploy.sh
@@ -29,21 +29,10 @@ check_update_needed () {
     # update the subchart version in sysdig-deploy/Chart.yaml
     yq -i '( .dependencies[] | select(.name == "'$chart'") | .version) = "~'"$new_subchart_version"'"' charts/$sysdig_deploy_path/Chart.yaml
 
-    for ((idx=0; idx<3; ++idx)); do
-        if [ ${next_ver[idx]} -gt ${prev_ver[idx]} ] 
-        then
-            if [ $idx -eq 0 ]
-            then
-                major=1
-            elif [ $idx -eq 1 ]
-            then
-                minor=1
-            elif [ $idx -eq 2 ]
-            then
-                patch=1
-            fi
-        fi
-    done
+    [ ${next_ver[0]} -gt ${prev_ver[0]} ] && major = 1
+    [ ${next_ver[1]} -gt ${prev_ver[1]} ] && minor = 1
+    [ ${next_ver[2]} -gt ${prev_ver[2]} ] && patch = 1
+
 }
 
 charts=( "$node_analyzer_chart_path" "$agent_chart_path" "$kspm_collector_chart_path" "$rapid_response_chart_path" "$admission_controller_chart_path" )
yash@yash charts % ./scripts/sysdig-deploy/update-sysdig-deploy.sh
subchart version in node-analyzer/Chart.yaml
1.7.33
node-analyzer subchart version in sysdig-deploy/Chart.yaml
1.7.33
```
the execution stops after the first chart

.github/workflows/update-sysdig-deploy-chart.yaml
iurly: > Hello for a) I can remove the paths from the workflow the only thing I can foresee is that it would run this workflow on any change made on this repository as a push to master, which should be fine since it won't make any changes. 

I'm not sure that would happen out of the box, unless you make the "won't make any changes" conditional on the script resulting in "no action necessary" -- that is the idea however.

> Let me know if that is preferable. 

That was my idea -- i.e., make the action idempotent -- but please don't feel any obligation to go for it if you don't agree.

> Regarding return 0 which means execution is a success is fine but if I return 1 shouldn't that lead to erroring out and the workflow not running successfully? 

True. Perhaps you can use something like this?
```
    run: |
      scripts/sysdig-deploy/update-sysdig-deploy.sh && SHORT_CIRCUIT=1 || true
```
so that when the script returns 0 (no difference), SHORT_CIRCUIT is set to 1 so to skip all the following steps.
Otherwise, it means the error (return 1) should be swallowed and execution should continue.
Notice the return values I suggested come from `man diff`:

```
Exit status is 0 if inputs are the same, 1 if different, 2 if trouble.
```

> b) I am assuming what is meant here is to refer to an environment variable after setting it once instead of the whole path. ill make that change now.

Yep, assuming it gets propagated correctly (which is what the comment I mentioned above suggests)

yashgiri: Hi @iurly I tried those changes and although it works fine when changes are present it still runs the subsequent changes when the changes are not present ("No modifications necessary"). I am more inclined towards keeping the changes currently present seeing that.

yashgiri:
Had a discussion with @iurly and after a successful review merging these changes.

PR #697, opened 2022-11-08
[sysdig-deploy] Add admission controller

ironashram:

- Add admission-controller as dependency of sysdig-deploy
- Add namespace field to all resources in admission-controller
- Fix README.md header navigation in sysdig-deploy
- Bump admission-controller version
- Bump sysdig-deploy version

## Checklist


lilx1ao (approved):

charts/sysdig-deploy/README.md
```diff
@@ -174,25 +174,59 @@ The following table lists the configurable parameters of this chart and their de
 | `global.clusterConfig.name`             | Identifier for this cluster                                                                                             | `""`      |
 | `global.sysdig.accessKey`               | Sysdig Agent Access Key                                                                                                 | `""`      |
 | `global.sysdig.accessKeySecret`         | The name of a Kubernetes secret containing an 'access-key' entry.                                                       | `""`      |
+| `global.sysdig.secureAPIToken`          | API Token to access Sysdig Secure                                                                                       | `""`      |
+| `global.sysdig.secureAPITokenSecret`    | The name of a Kubernetes secret containing API Token to access Sysdig Secure .                                          | `""`      |
```
aroberts87: nit: there is an extra space between 'Secure' and the ending period. Nearly every other description here doesn't end with a period, so we could very well just get rid of it.

charts/admission-controller/templates/_helpers.tpl
```diff
@@ -231,3 +238,43 @@ Allow overriding registry and repository for air-gapped environments
...
+{{- define "sysdig.secureAPIToken" -}}
+    {{- required "A valid secureAPIToken is required" (.Values.sysdig.secureAPIToken | default .Values.global.sysdig.secureAPIToken) -}}
+{{- end -}}
+
+{{- define "sysdig.secureAPITokenSecret" -}}
+    {{- .Values.sysdig.existingSecureAPITokenSecret | default .Values.global.sysdig.secureAPITokenSecret | default "" -}}
+{{- end -}}
+
+{{- define "clusterName" -}}
+    {{- required "A valid cluster name is required" (.Values.clusterName| default .Values.global.clusterConfig.name) -}}
```
aroberts87: super minor nit: all other parameters/values have a space on either side of the `|`, but you're missing one after `.Values.clusterName` here

aroberts87 (approved):
Two super minor nits. I'm approving since I'd be okay if you merged without fixing.

ironashram:
> Two super minor nits. I'm approving since I'd be okay if you merged without fixing.

thanks, fixed both

charts/admission-controller/templates/scanner/deployment.yaml
```diff
@@ -73,31 +74,31 @@ spec:
           env:
             {{- if (or .Values.scanner.httpProxy .Values.scanner.httpsProxy) }}
```
dark-vex: @ironashram I believe you should change also this line otherwise the proxy will get set only when using `admissionController.scanner.httpProxy` and gets ignored when using the global settings
cc: @lilx1ao @aroberts87

ironashram: indeed, great catch thank you

lilx1ao: nice!

dark-vex: Thanks for fixing it Michele! 

One more question, I see you have also set the namespace variable for multiple objects, like: secrets, service, role, PodSecurityPolicy, etc.. 

What's the purpose of that? You would like to give the possibility to deploy the admission controller on a different namespace while using `sysdig-deploy`? I'm asking because by default without specifying it, helm will take care of patching the manifest files and add the namespace specified with `helm install` command

charts/admission-controller/templates/scanner/psp.yaml
```diff
@@ -2,7 +2,8 @@
 apiVersion: policy/v1beta1
```
dark-vex, line 2: Don't hate me but could you please add the k8s version check to this PSP policy? Like what we did for the agent https://github.com/sysdiglabs/charts/blob/master/charts/agent/templates/psp.yaml#L1
Otherwise the deploy on Kubernetes v1.25 will fail due to the deprecation of PSP policies

dark-vex (changes requested):

yashgiri (approved):
Thank you for the changes 👍

charts/admission-controller/templates/scanner/psp.yaml
ironashram: done!

charts/admission-controller/templates/scanner/deployment.yaml
ironashram: the idea was to give that option yes, but also to try and align with the other charts which are already among the dependencies

charts/admission-controller/values.yaml
```diff
@@ -2,18 +2,31 @@
...
+  # Global HTTP Proxy settings.
+  proxy: {}
 
 # **required**
 # <br/>Cluster Name which appear on Secure UI
 clusterName: ""
 
-sysdig:
+# Namespace to install components, if not specified `Release.Namespace` is used.
+namespace: ""
```
dark-vex, line 23: I did a test by setting a different namespace for the admission controller and in case the namespace doesn't already exist the whole helm installation fails:
```
Error: INSTALLATION FAILED: namespaces "test" not found
```
this is the command used:
```
 ➜ helm install sysdig-agent -f values-test.yaml charts/sysdig-deploy -n sysdig-agent --create-namespace
```

values-test.yaml
```
global:
  clusterConfig:
    name: "test-cluster"
  sysdig:
    accessKeySecret: "sysdig-agent"
    region: "us1"
  image:
    registry: quay.io
  proxy:
    httpProxy: http://squid.mydomain.tld:3128
    httpsProxy: http://squid.mydomain.tld:3128
    noProxy: 100.96.0.0/10

admissionController:
  enabled: true
  namespace: "test"
  sysdig:
    secureAPIToken: test

nodeAnalyzer:
  nodeAnalyzer:
    benchmarkRunner:
      deploy: false
      resources:
        limits:
          cpu: 250m
    runtimeScanner:
      deploy: true
      eveConnector:
        deploy: true
      resources:
        limits:
          cpu: 300m
      settings:
        eveEnabled: true
  secure:
    vulnerabilityManagement:
      newEngineOnly: true
```

I don't like too much the idea of giving the possibility to specify a different namespace while using sysdig-deploy chart but seems to be fine for the other maintainers so we can keep it :) 
Just please ensure to add a note on the `readme.md` file that for the specified namespace, it must exist prior to apply/install the chart

dark-vex (commented):
Added a comment for the namespace flag, aside that, the PR LGTM

charts/admission-controller/values.yaml
lilx1ao: good point. this is a problem. may need to discuss with the product team before committing to this.

dark-vex (approved):

charts/admission-controller/values.yaml
ironashram: i have added a note for the namespace flag, it should be quite clear now

PR #702, opened 2022-11-09
[CI] Adding unit test workflow for agent and nodeanalyzer

yashgiri:

This PR is to add the workflow changes to add unit tests to the (Lint and test charts) workflow. Once this workflow is added the test suite in the agent and node-analyzer will always be activated and any further test changes will need to be done only in the test suites.

Also, the current tests being added are the ones checking collector endpoint and apiEndpoint values.

## Checklist

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -62,3 +62,24 @@ jobs:
 
       - name: Run chart-testing (install)
         run: ct install --upgrade
+  Unit-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      
+      - name: Set up Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.4.0
+
+      - name: Set up helm unit test plugin
+        run: helm plugin install https://github.com/quintush/helm-unittest
+      
+      - name: Test node-analyzer
+        run: helm unittest --helm3 ./charts/node-analyzer
+
+      - name: Test agent
+        run: helm unittest --helm3 ./charts/agent


yashgiri:
Hi please check out the Lint and Test Charts / Unit-tests (pull_request) to see its execution.

ironashram (commented):
should be ok, i was wondering one thing tho, do we need to initialize another job for this? or would it make sense to just add the steps to the already present lint-test job?, we would avoid the need to setup helm and checkout twice.

Another option would be to add the unit tests as a separate workflow
I don't have a strong opinion on this, just trying to think what would be the best going forward

yashgiri:
That's actually a great question and I did give it quite some thought.
So I kept it in this workflow by going through this: https://docs.github.com/en/actions/using-workflows/about-workflows
it felt that since it needs to be activated on the same event which is a pr it makes sense. 
The decision to keep it in a different job was purely based on reducing CI time and improving parallelization. As we go ahead a lot of Unit tests will be automated and added to this 'Unit-tests' job and the 'lint-test' takes quite some time.

But yeah if there is any other suggestion based on some reasoning that you think is better suited we can go that way. I also do not really have a strong opinion on this.

I will wait for @aroberts87 review on this

aroberts87 (approved):
> do we need to initialize another job for this?

I can see the argument for both sides of this discussion, but ultimately I feel that having a different job makes the most sense. As @yashgiri mentioned before, having separate jobs helps with operational logistics with parallelization and whatnot, but it also helps logically separate two different workflows (linting vs testing).

@yashgiri - I'm good with this PR as it stands today though, feel free to adjust if you want to based on feedback, but this LGTM.

yashgiri:
I squashed the commits and have rebased it

PR #708, opened 2022-11-11
[admission-controller] fix: Remove strict requirement for secureAPIToken

ironashram:
## What this PR does / why we need it:

Fix bug introduced with latest changes, it was not possible to use secureAPITokenSecret

## Checklist

[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]

- [x] Title of the PR starts with chart name (e.g. [mychartname])
- [x] Chart Version bumped for the respective charts
- [x] Changelog updated for the charts with version updates.
- [ ] Variables are documented in the README.md (or README.tpl in some charts)
- [x] Check GithubAction checks (like lint) to avoid merge-check stoppers
- [ ] All test files are added in the tests folder of their respective chart and have a "_test" suffix

Check Contribution guidelines in README.md for more insight.

PR #709, opened 2022-11-14
[sysdig-deploy][admission-controller] Add Secure API token validation for Admission Controller

ironashram:

- Add Secure API token validation for Admission Controller
- Bump sysdig-deploy and admission-controller

This PR will make specifying either a `secureAPIToken` or a `secureAPITokenSecret` a requirement and return a meaningful error
Mainly meant as an improvement over https://github.com/sysdiglabs/charts/pull/708

The only concern is that before adding the admission-controller to sysdig-deploy this requirement was not present and only a warning message was returned.

## Checklist

### DIFF
--- a/charts/admission-controller/templates/NOTES.txt
+++ b/charts/admission-controller/templates/NOTES.txt
@@ -1,4 +1,6 @@
+Validating Admission Controller Secure API Token configuration:
 {{- if (include "admissionController.validTokenConfig" .) }}
+Success!
 {{- end }}
 
 Sysdig Admission Controller is now installed!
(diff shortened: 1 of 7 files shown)


charts/admission-controller/templates/NOTES.txt
```diff
@@ -1,16 +1,12 @@
+{{- if (include "admissionController.validTokenConfig" .) }}
+{{- end }}
```
lilx1ao, line 4: what's this for?

ironashram: the `admissionController.validTokenConfig` is the named template in the `helpers.tpl` which validates the config, but since it's not a parameter used in the actual chart templates i added this "if" here in order to trigger the error message, otherwise the check would not work

Maybe i can add a comment to explain this

lilx1ao: interesting! I like it

lilx1ao (approved):

aroberts87 (approved):
LGTM! Thanks for this change!

yashgiri (approved):

PR #725, opened 2022-11-15
[CI] Updated sysdig deploy update workflow with new PAT token

yashgiri:
## Checklist


aroberts87:
@yashgiri - am I correct in assuming the 'P' in PAT stands for personal? In that case, what would happen to this workflow should you ever leave Sysdig?

yashgiri:
@aroberts87  The workflow would work as intended until Feb 13, 2023 which is the expiration date of the token. After which the creation of the PR would fail. At that point, the token might have to be replaced. I can also create a token with longer expiry if needed for 2-3 years or so.
https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/token-expiration-and-revocation

Edit: I realized irrespective of the token expiry date if the token holder's access is revoked the workflow would fail from the time the access is revoked.

aroberts87 (approved):
@yashgiri - i'm going to approve this now, but we need to find a better long term solution for how to acquire the token.

yashgiri:
The token has been approved merging this PR now.

PR #730, opened 2022-11-16
[agent][node-analyzer][sysdig-deploy] Update .helmignore files

aroberts87:
A previous change to the .helmignore accidentally caused the `$chart/templates/tests/` folders in some subcharts to be omitted, which prevented the use of `helm test` on our releases. This change adds them back.

## Checklist


yashgiri:
Hi @aroberts87 I just checked "/tests/" was not present in the .helmignore of the other 3 sub charts (rapid-response, kspm-collector, admission-controller) do you think it should be a part of this PR or should I create another PR for the same?

Also thank you for these changes!

yashgiri (approved):

lilx1ao:
does this pr have something to do with the test pod issue raised in one of our channels?

yashgiri:
I believe this PR modifies my changes to helmignore which unintentionally ignored the helm test that is to be included in their packaging.

lilx1ao:
thanks. so they wanted the test pod?

aroberts87:
> thanks. so they wanted the test pod?

no, they don't want the test pod. i'm still trying to figure out why they're getting it, but in my investigations i found out that we had a regression in the charts and were no longer shipping out chart tests. this commit resolves that issue.

aroberts87:
> Hi @aroberts87 I just checked "/tests/" was not present in the .helmignore of the other 3 sub charts (rapid-response, kspm-collector, admission-controller) do you think it should be a part of this PR or should I create another PR for the same?

I did not add to those files because the regression of ignoring just `tests` was not present in those `.helmignore` files. I'd be happy to make those changes in this PR if you'd like though.

lilx1ao (approved):
I see, thank you!

yashgiri:
@aroberts87 I think we can keep it in mind for the next time better to get this PR in right now.

PR #735, opened 2022-11-18
[agent] Add 'beta' tag back to agent DS

aroberts87:

The removal of 'beta' from the agent daemonset `kubernetes.io/arch` and `kubernetes.io/os` node selector broke OKD3 installs since OKD3 is based off of Kubernetes 1.11. This commit adds the 'beta' extension back at the expense of some warnings during installation of more modern clusters.

## Checklist

### DIFF
--- a/charts/agent/CHANGELOG.md
+++ b/charts/agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Change Log
 
+## v1.5.36
+### Minor changes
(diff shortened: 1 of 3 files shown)


yashgiri (approved):
The changes look good to me.
Just a suggestion - do you think a test could also be included for these changes in the unit tests? so this doesn't break the next time.
feel free to ignore It in case this change is time sensitive.

PR #740, opened 2022-11-21
Add greeting message on PRs from forks

ironashram:
Signed-off-by: Michele Palazzi <michele.palazzi@sysdig.com>

Adds a comment after a PR from fork is opened with useful info about what comes next

Example command usage https://github.com/sysdiglabs/charts/pull/739

## Checklist

### DIFF
--- a/.github/workflows/greeting.yaml
+++ b/.github/workflows/greeting.yaml
@@ -1,0 +1,27 @@
+name: Greeting
+on:
+  pull_request_target:
+    types:
+      - opened
+
+jobs:
+  greeting:
+    permissions:
+      pull-requests: write
+    name: Greeting comment upon PR from fork
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository
+
+    steps:
+      - name: Issue the greeting comment
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            Hi @${{ github.event.pull_request.user.login }}. Thanks for your PR.
+
+            After inspecting your changes someone with write access to this repo needs
+            to approve and run the workflow with the follwoing slash_command.
+
+            * `/approve-test-run sha=<short commit sha to test>`:            Run helm chart linting and testing workflow
--- a/.github/workflows/greeting.yaml
+++ b/.github/workflows/greeting.yaml
@@ -22,6 +22,6 @@ jobs:
             Hi @${{ github.event.pull_request.user.login }}. Thanks for your PR.
 
             After inspecting your changes someone with write access to this repo needs
-            to approve and run the workflow with the follwoing slash_command.
+            to approve and run the workflow with the following slash_command.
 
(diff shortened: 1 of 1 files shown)


.github/workflows/greeting.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Issue the greeting comment
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            Hi @${{ github.event.pull_request.user.login }}. Thanks for your PR.
+
+            After inspecting your changes someone with write access to this repo needs
+            to approve and run the workflow with the follwoing slash_command.
```
lilx1ao: following*

.github/workflows/greeting.yaml
```diff
@@ -0,0 +1,27 @@
...
+      pull-requests: write
+    name: Greeting comment upon PR from fork
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository
+
+    steps:
+      - name: Issue the greeting comment
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
lilx1ao, line 19: is there an org token we can replace this generic token with?

.github/workflows/greeting.yaml
```diff
@@ -0,0 +1,27 @@
...
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            Hi @${{ github.event.pull_request.user.login }}. Thanks for your PR.
+
+            After inspecting your changes someone with write access to this repo needs
+            to approve and run the workflow with the following slash_command.
+
+            * `/approve-test-run sha=<short commit sha to test>`:            Run helm chart linting and testing workflow
```
lilx1ao, line 27: is this comment a suggestion?

ironashram: This is what we ( maintainer/code owner) need to do in order to run the workflow in the PR from the fork, but i guess it's not clear

.github/workflows/greeting.yaml
ironashram: i could use the same one that gets used to bump the sysdig-deploy version, but it's called `DEPLOY_VERSION_CHANGE_TOKEN` which is not what we are doing here

or request a new one?  seems overkill tbh

.github/workflows/greeting.yaml
ironashram: this whole greeting idea was just to make things clearer and avoid having to search in weird places what to do in order to run the workflow, but we can also avoid it entirely

.github/workflows/greeting.yaml
lilx1ao: I see. I don't feel strongly either way.

.github/workflows/greeting.yaml
ironashram: For context, if no one runs the workflow with the `/approve-test-run` an approved PR could end up merged without the check, because by default it is skipped

lilx1ao: I love the greetings!

lilx1ao (approved):

PR #754, opened 2022-11-23
new: add automated changelogs

ironashram:
Signed-off-by: Michele Palazzi <michele.palazzi@sysdig.com>

Automate change log and release notes based on commit messages

The decision to use https://github.com/git-chglog/git-chglog comes from the fact that we have a monorepo with many helm charts and a lot of the other tools/actions available do not handle this nicely or at all.

The auto-generated `CHANGELOG.md` and `RELEASE-NOTES.md` will be pushed to master after the `release-charts ` step is successfull (these CI commits will not show up in the change logs as they are being ignored)

In order to do it i had to fork `git-chglog` and do some minor changes, i will try to upstream them but the project does not seem particularly active lately, this could be the only concern, anyway  this is something that we can easily replace if a better option is discovered in the future.

~~The manually updated `CHANGELOGS.md` will be renamed to `old_CHANGELOGS.md`  and will be referenced in the newly generated ones in order to preserve them.~~

Reworked based on @iurly suggestions to only generate the changelog for last release (tag).
The generated entry will then be inserted into the existing file after the '## Change Log' line.
This also allows to manually edit/fix changelogs if needed.

Example fake release screenshot
Screenshot_20221123_180522

## Checklist


.chglog/config-release.yml
```diff
@@ -0,0 +1,36 @@
...
+    filters:
+      Type:
+        - feat
+        - new
+        - fix
+        - perf
+        - refactor
+        - chore
+        - docs
+        - github_actions_ci
```
yashgiri, line 15: I am assuming from here on the commit messages would have to include these filters as prefixes. This will require telling every person who commits to have that tag present correct?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,11 +4,10 @@
 
 [Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
 
-- [ ] Title of the PR starts with chart name (e.g. [mychartname])
+- [ ] Title of the PR starts with type and scope, type must be one of feat, new, fix, perf, refactor, chore, docs (e.g. fix(node-analyzer):)
```
yashgiri: Considering we want every PR/commit to have type & scope as a prefix in the Changelog maybe we can make it mandatory in the lint-test (pull-request based workflow) for every PR to have that. what do you think?
Also, this example can be made more descriptive like what would a fix for 2 charts agent and rapid-response look like. i am assuming fix(agent,rapid-response):

ironashram: The match is done via regex, this is an example https://regex101.com/r/WbpM78/1

As it stands `fix(agent,rapid-response):` would not work, we can play around with the regex to include something like that, anyway i'm not a huge fan of the extremely long and unreadable commit messages we have right now
My idea would be that if your change affects more than one chart the "scope" could be something like global, or all (e.g. new(global): add automated changelogs)

.chglog/config-release.yml
ironashram: If the commit does not match the regex pattern it will be listed in the "Others" section and will not have as much details
So yes ideally every commit should adhere to these standards

.github/PULL_REQUEST_TEMPLATE.md
ironashram: I agree on making it mandatory in the PR, we should be able to easily check that

ironashram: fyi improved the regex a bit, now `fix(agent,rapid-response):` is allowed, also made a few `README` adjustments

iurly (approved):
LGTM!
My only concern is about the changelog being perpetually overwritten (and hence, the existing one having to be renamed).
I'd expect there to be a way for the automation to only produce the changelog for the just-added commits, and only prepend that into an existing changelog whose contents is otherwise arbitrary. Did you look into that?

ironashram:
> LGTM! My only concern is about the changelog being perpetually overwritten (and hence, the existing one having to be renamed). I'd expect there to be a way for the automation to only produce the changelog for the just-added commits, and only prepend that into an existing changelog whose contents is otherwise arbitrary. Did you look into that?

The automation will produce a changelog only for the commits since the last tag of that particular chart, the problem is more for the first time, since it will do for all tags

There is an option to exclude tags from the changelog generation, but it has to specified in the [template](https://github.com/sysdiglabs/charts/blob/feature/automated-changelogs/.chglog/CHANGELOG.tpl.md) and this means we would need to exclude all the existing tags(which are a huge amount and sprig is not that flexible) there and have a specific template for every chart, which seems a lot worse than this to me, anyway i will have another look maybe there is some other way

ironashram:
reworked the whole thing a bit, now it does generate only the latest changelog and also allows manual edits, i will need to unify the Changelog Header for all the charts for this to work properly but should not be an issue

i was thinking something like

```
# Chart: chart-name

All notable changes to this chart will be documented in this file.

The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).

Please note that it's automatically updated vía github actions.
Manual edits are supported only below '## Change Log' and should be used 
exclusively to fix incorrect entries and not to add new ones.

## Change Log
```

aroberts87 (approved):

yashgiri (approved):
LGTM thank you for these changes

aroberts87:
Something I just thought about, @ironashram . This change will likely require a Slack blast to #agent-helm-refactor to notify about the new PR title requirements. I could see this getting push back if this goes live without prior notice. You may even want to put out the notice tomorrow (Friday) and hold off on the merge until next week to give people time to read it. Just my opinion though.

README.md
```diff
@@ -22,16 +22,38 @@ If you make changes to an existing chart, but do not change its version, nothing
...
+- chore
+- docs
+
+A Scope should be always present, a few examples:
+- (agent)
+- (sysdig-deploy)
+- (agent,node-analyzer,sysdig-deploy)
+
+Full PR title example
+`feat(agent,node-analyzer,sysdig-deploy): add automated changelogs`
```
guidobonomi, line 47: we've been asked to include, as part of the changelog, elaborated content like the list of the vulnerabilities addressed with a particular change. Example here: https://github.com/sysdiglabs/charts/pull/766/files#diff-a465eea3af5b1cf427e17ce8720fc5bb4e1136c508c2954675599a63961bde21R7
Could you please specify what would be the proper way to generate something like that in the Changelog?

ProvoK: I've use git-chlog in the past and AFAIK it's possible to fetch the commit <body> for this kind of stuff.

git commit body is basically everything other than the first line, and could help here @ironashram

.chglog/CHANGELOG.tpl.md
```diff
@@ -0,0 +1,22 @@
+{{ if .Versions -}}
+{{ range .Versions -}}
+{{- $chartName := regexReplaceAll "((?:[A-Za-z]*(?:\\-[A-Za-z]*)*))\\-(?:.*)" .Tag.Name "${1}" -}}
+{{- $chartversion := regexReplaceAll "(?:[A-Za-z]*(?:\\-[A-Za-z]*)*)\\-(.*)" .Tag.Name "v${1}" -}}
+# {{ $chartversion }}
+{{ range .CommitGroups -}}
+{{- if not (eq "Ignored" .Title ) -}}
+### {{ .Title }}
+{{ range .Commits -}}
+* {{ if .Scope }}**{{ .Scope }}** [{{ end }}[{{.Hash.Short}}]({{ $.Info.RepositoryURL }}/commit/{{ .Hash.Long }})]: {{ .Subject }}
```
ProvoK, line 10: AFAICT from the regex below, `.Scope` will be the same for all charts, causing some weirdness in the changelogs.
image

eg: commit `fix(node-analyzer,agent): something`, will `(node-analyzer,agent)`  be present in each `agent` and `node-analyzer` changelogs? 🤔 


We could fix this in two ways IMVHO:
1) Making the regex more complex, and grouping differently every substring in the `(...)`, comma separated, and match the chart name for the appropriate changelog message
2) Don't allow multi-chart changes 😄 Which makes probably more sense to me, given each chart will actually have slightly different changes (`sysdig-deploy` is usually just bumps in versions)


...or even totally ignore the weirdness and leave as it is - but I would rather choose one of the above.

If you have any other idea, please share @ironashram 🙏

README.md
ironashram: indeed, thanks, i am going to look into how to add it properly

.chglog/config-changelog.yml
```diff
@@ -0,0 +1,33 @@
...
+  title: CHANGELOG
+  repository_url: https://github.com/sysdiglabs/charts
+options:
+  commits:
+    filters:
+      Type:
+        - feat
+        - new
+        - fix
+        - perf
```
ProvoK:
```suggestion
        - fix
```

very minor, but I believe we don't need `new` (when use `new` and when `feat` ?) and I can't think about `perf` use-cases.

Again, very minor comments, just trying to simplify the <what-should-I-use> dilemma when opening a PR 😄

ProvoK: (same consideration applies for the config-release.yml below)

ProvoK (commented):
thanks a LOT for this @ironashram 🙏 left a couple of comments

ProvoK:
last but not least, and maybe not in scope for this PR, but maybe we could handle the semver bump (Chart.yaml) automatically too, given we know the <kind> of the change

.chglog/CHANGELOG.tpl.md
ironashram: @ProvoK My initial idea was exactly your option 2, then @yashgiri  requested to support the multi chart `Scope`

The option 3 is that if the change affects multiple charts the `Scope` could be either avoided entirely or something generic in order to avoid weirdness in the changelogs

The thing that i would really like to avoid is over complicating the logic and the regex

ProvoK: > The thing that i would really like to avoid is over complicating the logic and the regex

totally agree with this. Our life is hard enough even without regexes 🤣

README.md
ironashram: fyi i added a way to add an extendend changelog entry and also a dedicated section with an example in the `README`, it is not perfect since one will need to remember to add this section while squashing and merging, but i haven't found a better way

ironashram:
> last but not least, and maybe not in scope for this PR, but maybe we could handle the semver bump (Chart.yaml) automatically too, given we know the of the change

indeed, but i would leave that for a separate PR since this is already a lot

ProvoK (approved):
LGTM thanks a lot for this @ironashram !

PR #772, opened 2022-11-29
[sysdig-deploy][agent] Fixed notes section links and added tests

yashgiri:
## Checklist


charts/agent/tests/notes_test.yaml
```diff
@@ -0,0 +1,74 @@
...
+          pattern: https://app.au1.sysdig.com/secure/#/data-sources/agents
+
+  - it: Checking incorrect region 'ap3' should fail
+    set:
+      global: 
+        sysdig:
+          region: ap3
+    asserts:
+      - failedTemplate:
+          errorMessage: "raw: global.sysdig.region=ap3 provided is not recognized."
```
ironashram: newline

charts/sysdig-deploy/tests/notes_test.yaml
```diff
@@ -0,0 +1,74 @@
...
+          pattern: https://app.au1.sysdig.com/secure/#/data-sources/agents
+
+  - it: Checking incorrect region 'ap3' should fail
+    set:
+      global: 
+        sysdig:
+          region: ap3
+    asserts:
+      - failedTemplate:
+          errorMessage: "raw: global.sysdig.region=ap3 provided is not recognized."
```
ironashram: same as above

ironashram (approved):
looks good, except two newlines missing and the conflicts to resolve

charts/agent/templates/_helpers.tpl
```diff
@@ -242,6 +242,52 @@ Determine collector endpoint based on provided region
...
+        {{- fail (printf "global.sysdig.region=%s provided is not recognized." .Values.global.sysdig.region ) -}}
+    {{- end -}}
+{{- end -}}
+
+{{/*
+Determine sysdig secure endpoint based on provided region
+*/}}
+{{- define "secureUrl" -}}
+    {{- if (or .Values.collectorSettings.collectorHost (eq .Values.global.sysdig.region "custom")) -}}
+        {{- required "collectorSettings.collectorHost is required for custom regions" (.Values.collectorSettings.collectorHost) -}}
```
aroberts87, line 273: Secure also has a few places where we need to specify `apiEndpoint` as well, like https://github.com/sysdiglabs/charts/blob/master/charts/node-analyzer/values.yaml#L86. Should we make those a requirement here too?

yashgiri: I see it as it should be done if <that condition> then make it required. It's no use pushing a requirement when it is not needed. For this particular case in notes it makes sense to add that requirement only if it will be used to generate the secureUrl that will be presented in the notes section. if the apiEndpoint needs to be at some other place it should be at that particular place.

aroberts87 (approved):

PR #779, opened 2022-12-01
[sysdig-deploy] Fixed custom option breaking notes section

yashgiri:
## Checklist

### DIFF
--- a/charts/sysdig-deploy/templates/_helpers.tpl
+++ b/charts/sysdig-deploy/templates/_helpers.tpl
@@ -15,7 +15,9 @@ Determine sysdig monitor endpoint based on provided region
     {{- else if (eq .Values.global.sysdig.region "au1") -}}
         {{- "app.au1.sysdig.com" -}}
     {{- else -}}
-        {{- fail (printf "global.sysdig.region=%s provided is not recognized." .Values.global.sysdig.region ) -}}
+        {{- if (ne .Values.global.sysdig.region "custom") -}}
+            {{- fail (printf "global.sysdig.region=%s provided is not recognized." .Values.global.sysdig.region ) -}}
+        {{- end -}}
     {{- end -}}
 {{- end -}}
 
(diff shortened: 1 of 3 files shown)


charts/sysdig-deploy/templates/_helpers.tpl
```diff
@@ -14,8 +14,6 @@ Determine sysdig monitor endpoint based on provided region
         {{- "eu1.app.sysdig.com" -}}
     {{- else if (eq .Values.global.sysdig.region "au1") -}}
         {{- "app.au1.sysdig.com" -}}
-    {{- else -}}
-        {{- fail (printf "global.sysdig.region=%s provided is not recognized." .Values.global.sysdig.region ) -}}
```
aroberts87: Could we instead do a check for a custom region and rather than validate against a known URL, just print out a message like `Custom region has been specified`. That way we could still have a failure case if an invalid region was provided.

charts/sysdig-deploy/templates/_helpers.tpl
```diff
@@ -35,7 +33,5 @@ Determine sysdig secure endpoint based on provided region
         {{- "eu1.app.sysdig.com/secure" -}}
     {{- else if (eq .Values.global.sysdig.region "au1") -}}
         {{- "app.au1.sysdig.com/secure" -}}
-    {{- else -}}
-        {{- fail (printf "global.sysdig.region=%s provided is not recognized." .Values.global.sysdig.region ) -}}
```
aroberts87: Same comment here as above. Could we add an if clause for custom region and just print out that we've been given a custom region?

charts/sysdig-deploy/tests/notes_test.yaml
```diff
@@ -64,11 +64,3 @@ tests:
...
           pattern: https://app.au1.sysdig.com/secure/#/data-sources/agents
 
-  - it: Checking incorrect region 'ap3' should fail
-    set:
-      global: 
-        sysdig:
-          region: ap3
-    asserts:
-      - failedTemplate:
-          errorMessage: "raw: global.sysdig.region=ap3 provided is not recognized."
```
aroberts87: If we make the adjustments as mentioned above, we can add this test back.

aroberts87 (changes requested):

aroberts87 (approved):

PR #786, opened 2022-12-08
[node-analyzer]add namespace field [admission-controller] add option to set priorityClassName

a1ex-var1amov:

The PR includes:

- an option to define `namespace` for node-analyzer by default
- an option to set the `priorityClassName` for AC pods (webhook and scanner)

Tested with `ct` (aka `chart-testing`)

## Checklist


yashgiri (approved):
LGTM.

charts/node-analyzer/templates/_helpers.tpl
```diff
@@ -269,3 +269,10 @@ Deploy on GKE autopilot
         true
     {{- end -}}
 {{- end -}}
+
+{{/*
+Returns the namespace for installing components
+*/}}
+{{- define "nodeAnalyzer.namespace" -}}
+    {{- coalesce .Values.namespace .Values.global.clusterConfig.namespace .Release.Namespace -}}
+{{- end -}}
```
ironashram: please add a newline

ironashram (changes requested):
the AC chart uses `chart-doc-gen` to generate the `README`
please run `make docs` and commit the generated changes, otherwise when i trigger the CI workflow it will fail

ironashram (approved):
lgtm now, the PR title could be improved but that's not a blocker for now, it will be after https://github.com/sysdiglabs/charts/pull/754

ct.yaml
```diff
@@ -1,3 +0,0 @@
-remote: origin
-chart-dirs:
-    - charts
```
aroberts87, line 3: Just curious, why was this removed?

a1ex-var1amov: I think the ct util does it automatically; should we put it back?

regiluze (approved):

PR #800, opened 2022-12-12
fix(kspm-collector): Fix numeric port on configmap

chen-shmilovich-sysdig:
A fix for:
* Agent port configuration was applied when the `noProxy` variable was set.
* Wrap numeric agent port configuration value in quotes.

## Checklist


ironashram:
@chen-shmilovich-sysdig adding a changelog entry manually is no longer needed, when the PR title follows https://github.com/sysdiglabs/charts/blob/master/README.md#pull-requests an entry will be automatically generated

in this case you could use
`fix(kspm-collector): Fix numeric port on configmap`

aviad-chen-apolicy (approved):

PR #811, opened 2022-12-13
fix(sysdig-deploy): Chart documentation fix values missing from configuration but referenced in examples

yilmi:
People looking for the first time at the chart have a hard time understanding what gets deployed and not. The helm commands examples contain values that are not listed in the sysdig-deploy chart.

To help people understand how they can enable different components, I added some of them to the sysdig-deploy chart documentation.

I also direct users to the different sub-charts composing the sysdig-deploy chart so new folks know where to look for advanced config.

## Checklist

- [N/A] Chart Version bumped for the respective charts
- [N/A] All test files are added in the tests folder of their respective chart and have a "_test" suffix


ironashram:
@yilmi if the PR comes from a fork of the repo like this, a maintainer needs to check it and allow the tests to run, this is for security reasons

ironashram (changes requested):
```
 ✖︎ sysdig-deploy => (version: "1.5.17", path: "charts/sysdig-deploy") > Chart version not ok. Needs a version bump!
------------------------------------------------------------------------------------------------------------------------
Error linting charts: Error processing charts
Error: Process completed with exit code 1.
```
you need to bump the chart version, i will trigger the tests again after it

yilmi:
> ```
>  ✖︎ sysdig-deploy => (version: "1.5.17", path: "charts/sysdig-deploy") > Chart version not ok. Needs a version bump!
> ------------------------------------------------------------------------------------------------------------------------
> Error linting charts: Error processing charts
> Error: Process completed with exit code 1.
> ```
> 
> you need to bump the chart version, i will trigger the tests again after it

Thanks @ironashram, apologies I assumed version changes weren't needed for documentations changes.

I just bumped it, should be fine now 👍

charts/sysdig-deploy/README.md
```diff
@@ -168,7 +168,7 @@ helm upgrade -n sysdig-agent sysdig sysdig/sysdig-deploy -f values.sysdig.yaml
 
 ## Configuration
 
-The following table lists the configurable parameters of this chart and their default values.
+The following table lists the configurable parameters of this chart and their default values, it is indicate and is not complete. If you are looking for more advanced configuration values, those are available in the documentation of the admission-controller, agent, node-analyzer, kspm-collector and rapid-response charts.
```
ironashram:
```suggestion
The following table lists the configurable parameters of this chart and their default values, it is indicative and contains only the most commonly used parameters. If you are looking for more advanced configuration values, those are available in the documentation of the admission-controller, agent, node-analyzer, kspm-collector and rapid-response charts.
```

ironashram (changes requested):
i think it's a good addition, i suggested a minor rephrasing

would also like a review from @aroberts87

charts/sysdig-deploy/README.md
```diff
@@ -192,7 +192,11 @@ The following table lists the configurable parameters of this chart and their de
...
 | `nodeAnalyzer.enabled`                  | Enable the nodeAnalyzer component in this chart                                                                         | `true`    |
+| `nodeAnalyzer.secure.enabled`           | Enable Sysdig Secure                                                                                                    | `true`    |
+| `nodeAnalyzer.secure.vulnerabilityManagement.newEngineOnly` | Enable only the new vulnerability management engine                                                 | `false`   |
 | `nodeAnalyzer.nodeAnalyzer.apiEndpoint` | nodeAnalyzer apiEndpoint                                                                                                | `""`      |
+| `nodeAnalyzer.nodeAnalyzer.benchmarkRunner.deploy` | Deploy the Benchmark Runner Scanner                                                                          | `true`   |
+| `nodeAnalyzer.nodeAnalyzer.runtimeScanner.deploy` | Deploy the Runtime Scanner                                                                                    | `false`   |
```
lilx1ao, line 200: I remember Christian gave a presentation where some of these components require other values to be enabled or dismissed in order to work. @claffin

charts/sysdig-deploy/README.md
```diff
@@ -168,7 +168,7 @@ helm upgrade -n sysdig-agent sysdig sysdig/sysdig-deploy -f values.sysdig.yaml
 
 ## Configuration
 
-The following table lists the configurable parameters of this chart and their default values.
+The following table lists the configurable parameters of this chart and their default values, it is indicative and contains only the most commonly used parameters. If you are looking for more advanced configuration values, those are available in the documentation of the admission-controller, agent, node-analyzer, kspm-collector and rapid-response charts.
```
aroberts87: If I recall correctly, the only parameters/values we should be putting in the `sysdig-deploy` chart are those that can be used by multiple charts, or a simple boolean to enable/disable a given subchart. As such, I would recommend rephrasing this whole block similar to the following:

The following table lists the configurable parameters of the sysdig-deploy chart and their default values.  The sysdig-deploy chart itself only has select parameters that are used by multiple subcharts, and those used to enable/disable select subcharts. If you need additional configuration values, those are available in the various READMEs of the individual subcharts.

If you wanted to get fancy you could add links to the existing READMEs too.

charts/sysdig-deploy/README.md
aroberts87: Based on the precedent of the chart at it stands, I don't believe these parameters should be documented as a part of the `sysdig-deploy` chart. It is my understanding that the sysdig-deploy chart is only supposed to contain values that can be used by multiple subcharts ,and the simple boolean flags to enable/disable various subcharts. Any subchart specific configuration needs to be contained, and thus documented, within the subchart itself to conform with Helm parent/subchart best practices: https://helm.sh/docs/chart_template_guide/subcharts_and_globals/.

aroberts87 (dismissed):

charts/sysdig-deploy/README.md
yilmi: Hi @aroberts87 I understand the intent but the current state of the charts implementation doesn't allow it imho. The example commands we have in the readme do specifically reference those values, in which case it doesn't make much sense to not have them documented as well.

From our experience, we had to jump between the configuration of the different sub-charts and sometimes go inside the values definition and templates to understand how we could fix issues we had. I think Sysdig's users would benefit from more details, especially as more and more use only the sysdig-deploy chart and won't deploy other components individually.

The helm page about subcharts and globals, doesn't necessarily says that you shouldn't use subchart config:
> A parent chart can override values for subcharts.

For what it's worth, there is an automated way to generate the readme from the values file and then make sure values for subcharts are synced with the parent chart. 

Here is an example that bitnami is using: https://github.com/bitnami-labs/readme-generator-for-helm

charts/sysdig-deploy/README.md
yilmi: Thanks, I included your suggestion and added links as well :)

charts/sysdig-deploy/README.md
aroberts87: I certainly understand your points @yilmi, and agree that managing these values and documentation about them while spanning multiple teams is troublesome. IMHO the fact that we have to specify subchart overrides in the base `sysdig-deploy` installation instructions itself is an issue though and we should first look there. That said, the reason I'm being a bit of a stick in the mud about this is because of the ongoing effort by @claffin to reduce confusion in the helm docs. I'll defer to his call here.

yilmi: @aroberts87 thanks for taking the time to provide some more background as I wasn't aware of this work. I get that you may want to offer a better experience with global or high-level values for the main chart. I'd be happy to share more details about the challenges we faced, see if that could help you evaluate the relevance of some values.

claffin:
Thank you for the contribution @yilmi. 

I think this makes sense for now to have those configuration options explained in sysdig-deploy chart level, so customers can easily see why we're advising these defaults to be overwritten from the get go. I agree it goes against best practice for Helm, but until we improve the Helm chart further, I favour clarity for customers over best practice. 

I agree we should look to generate the readme from the values file, we can add this to our backlog.

charts/sysdig-deploy/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sysdig-deploy
 description: A chart with various Sysdig components for Kubernetes
 type: application
-version: 1.5.17
+version: 1.5.18
```
ironashram:
```suggestion
version: 1.5.20
```

ironashram (approved):
lgtm!, there were other PRs merged in the meantime so it just needs a newer version now

charts/sysdig-deploy/Chart.yaml
yilmi: Thanks, done

yilmi:
> Thank you for the contribution @yilmi.
> 
> I think this makes sense for now to have those configuration options explained in sysdig-deploy chart level, so customers can easily see why we're advising these defaults to be overwritten from the get go. I agree it goes against best practice for Helm, but until we improve the Helm chart further, I favour clarity for customers over best practice.
> 
> I agree we should look to generate the readme from the values file, we can add this to our backlog.

@claffin no worries, I just wanted to share one last piece of feedback about how far one should go hiding sub-chart values and what is the actual best practice. 

What Helm says is: 
> A subchart is considered "stand-alone", which means a subchart can never explicitly depend on its parent chart.

I read this as, the sub-chart should be deployable on its own (as an external example, take postgresql [here](https://github.com/bitnami/charts/blob/main/bitnami/postgresql/README.md))

> For that reason, a subchart cannot access the values of its parent.

To be standalone, the sub-chart shouldn't rely on its parent to provide global values, so they should also be defined in the subchart values, same example with postgresql [here](https://github.com/bitnami/charts/blob/main/bitnami/postgresql/README.md#global-parameters)

> A parent chart can override values for subcharts.

I read this as, while we have global values, we can't handle all the sub-chart values through global values (except for simple charts). As such, we want to still be able to override sub-chart values from the parent, without having to set a global value for it.

> Helm has a concept of global values that can be accessed by all charts.

Based on the example given in the helm documentation, I read this as global values should cover configuration that is "factorizable" or high level.

For example:
- global.benchmarkRunner.enabled or global.runtimeScanner.enabled should definitely exist
- However, some sub-chart specific configuration may still be needed, even at the parent chart level. A good example for that is the [airflow chart](https://github.com/bitnami/charts/tree/main/bitnami/airflow#parameters), that depends on the postgresql chart.
- An example specific to the node-analyzer has [some config](https://github.com/sysdiglabs/charts/blob/master/charts/node-analyzer/values.yaml#L275-L300) that could be made available in the parent documentation 

One case that did cause issues on our end was the fine tuning of resources for the sub-charts, and understanding how much resources are required to run all the components. For this reason, I think that in your case, you will also need to bring to the parent chart values documentartion, the sub-chart pods specific resources defaults values and to make it easier for us to override them if necessary.

A lot of this complexity could also be removed by the use of an operator, but I assume this will take a while.

I hope you this is helpful, feel free to reach out directly if you want to chat further.

Thanks

PR #821, opened 2022-12-16
fix(sysdig,agent): add extra secrets and tests

jsecchiero:

* replaced '.' with '$' in secret range loop otherwise context is not passed to helm define functions in sysdig chart
* added helm unit tests for 'extraSecrets' in agent chart

## Checklist

### DIFF
--- a/charts/agent/templates/secrets.yaml
+++ b/charts/agent/templates/secrets.yaml
@@ -20,6 +20,8 @@ metadata:
   labels:
 {{ include "agent.labels" $ | indent 4 }}
 type: Opaque
-data: {{ toYaml .data | nindent 2 }}
+data:
+{{ toYaml .data | indent 2 }}
 ---
 {{- end }}
+
--- a/charts/sysdig/templates/secrets.yaml
+++ b/charts/sysdig/templates/secrets.yaml
@@ -14,10 +14,12 @@ data:
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ template "sysdig.fullname" . }}
+  name: {{ .name }}
   labels:
-{{ include "sysdig.labels" . | indent 4 }}
+{{ include "sysdig.labels" $ | indent 4 }}
 type: Opaque
-data: {{ toYaml .data | nindent 2 }}
+data:
+{{ toYaml .data | indent 2 }}
 ---
 {{- end }}
+
--- a/charts/agent/tests/secrets_test.yaml
+++ b/charts/agent/tests/secrets_test.yaml
@@ -1,0 +1,56 @@
+suite: Test agent secrets
+templates:
+  - templates/secrets.yaml
+tests:
+  - it: Checking default values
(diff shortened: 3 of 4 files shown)


ironashram:
approve-test-run sha=9f1dc30

ironashram (approved):
thanks for the quick fix! :rocket:

charts/sysdig/templates/secrets.yaml
```diff
@@ -14,9 +14,9 @@ data:
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ template "sysdig.fullname" . }}
+  name: {{ .name }}
   labels:
-{{ include "sysdig.labels" . | indent 4 }}
+{{ include "sysdig.labels" $ | indent 4 }}
```
lilx1ao, line 19: question: why is the dollar sign preferred?

jsecchiero: inside a range you change 'context' and '.' is the relative path to the current context.
with '$' you are specifying to pass all the values from the root

charts/sysdig/templates/secrets.yaml
```diff
@@ -14,9 +14,9 @@ data:
...
 type: Opaque
 data: {{ toYaml .data | nindent 2 }}
```
lilx1ao: I didn't get to comment on your previous pr, but now I have a chance :) 
I don't think we use nindent often, maybe its better if we keep it constant?

charts/sysdig/templates/secrets.yaml
lilx1ao: interesting. good to know, thank you!

aroberts87:
FWIW - @jsecchiero, you could write some unit tests for this change like some of the other charts are already doing. ex. https://github.com/sysdiglabs/charts/blob/master/charts/agent/tests/log_priority_test.yaml. It may help you catch some of these issues earlier next time, or more importantly make sure things you add don't get broken accidentally by someone else. 🙂  The docs for the unittest module we use are here: https://github.com/quintush/helm-unittest/blob/master/DOCUMENT.md

lilx1ao (approved):

lilx1ao (approved):

charts/agent/templates/secrets.yaml
```diff
@@ -20,6 +20,8 @@ metadata:
   labels:
 {{ include "agent.labels" $ | indent 4 }}
 type: Opaque
-data: {{ toYaml .data | nindent 2 }}
+data:
+{{ toYaml .data | indent 2 }}
```
lilx1ao, line 24: I didn't know the agent chart was doing it this way. my bad - thanks tho

jsecchiero:
@aroberts87 thanks for the suggestion. Tests added for agent chart for 'extraSecrets'

aroberts87 (approved):

PR #858, opened 2022-12-22
fix(agent): Fix agent.secure.enabled flag

aroberts87:
Setting `agent.secure.enabled=false` was not explicitly disabling the secure features in the generated agent configuration. Instead, the behavior was left up to the agent process's defaults, or a backend config push. This behavior is not desirable as a user who explicitly set `agent.secure.enabled=false` could have ended up with secure components running anyway. This fix ensures that setting `agent.secure.enabled=false` explicitly generates the configuration entries to force all secure components to be disabled.

## Checklist

### DIFF
--- a/charts/agent/templates/_helpers.tpl
+++ b/charts/agent/templates/_helpers.tpl
@@ -318,3 +318,31 @@ Use global sysdig tags for agent
         {{- end -}}
     {{- end -}}
 {{- end -}}
+
+{{/*
+Determine Sysdig Secure features that need to be enabled/disabled
+*/}}
+{{- define "agent.secureFeatures" }}
+    {{- $secureBlockConfig := dict "security" (dict
+        "enabled" .Values.secure.enabled
+        "k8s_audit_server_enabled" .Values.auditLog.enabled) }}
+    {{- if .Values.auditLog.enabled }}
+        {{- range $key, $val := (dict
+                 "k8s_audit_server_url" .Values.auditLog.auditServerUrl
+                 "k8s_audit_server_port" .Values.auditLog.auditServerPort) }}
+            {{- $_ := set $secureBlockConfig.security $key $val }}
+        {{- end }}
+    {{- end }}
+    {{- if not .Values.secure.enabled }}
+        {{- range $secureFeature := (list
+            "commandlines_capture"
+            "drift_killer"
+            "falcobaseline"
+            "memdump"
+            "network_topology"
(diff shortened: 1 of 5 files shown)


charts/agent/templates/configmap.yaml
```diff
@@ -13,20 +13,10 @@ data:
...
-  {{- end }}
-  {{- if .Values.secure.enabled }}
-      enabled: true
-    commandlines_capture:
-      enabled: true
-    memdump:
-      enabled: true
-  {{- end }}
+{{- if and (not .Values.secure.enabled) .Values.auditLog.enabled }}
+  {{- fail "auditLog.enabled cannot be true when secure.enabled is false" }}
```
lilx1ao: tiny suggestion: might make sense to do it from within the agent.secureFeatures?

aroberts87: Sure, I can do this. I think I tried that before and ended up with weird blank lines that I couldn't solve with adding `-`'s, but I agree with you it would look cleaner in the template. Let me try again. 💪🏻

lilx1ao: 💪

aroberts87: Done!

lilx1ao (approved):

charts/agent/templates/_helpers.tpl
```diff
@@ -318,3 +318,31 @@ Use global sysdig tags for agent
         {{- end -}}
     {{- end -}}
 {{- end -}}
+
+{{/*
+Determine Sysdig Secure features that need to be enabled/disabled
+*/}}
+{{- define "agent.secureFeatures" }}
+    {{- $secureBlockConfig := dict "security" (dict
+        "enabled" .Values.secure.enabled
```
balinaveen, line 337: May want to explain why secure.enabled = true and false are not mirror images.
For secure.enabled, only security.enabled is set to true (along with the k8s_audit_server_enabled set to whatever the auditLog.enabled is)
For secure.enabled=false, all known secure features are disabled.

The explanation is that for secure.enabled=true we depend on the backend to enable the appropriate secure features for the customer. And since we don't know a priori what secure features may be enabled when secure.enabled=true we disable them all when secure.enabled=false

aroberts87: Done!

balinaveen: Thanks

PR #862, opened 2022-12-29
fix(agent): Configure GOMAXPROCS to reduce CPU throttling

stevo-f3:
Signed-off-by: Stevo Slavić <stevo.slavic@form3.tech>

We're evaluating sysdig, and sysdig container in agent DaemonSet is misbehaving, CPUThrottlingHigh is alerting for it - even though it has several resource profiles defined all with CPU limits, Go runtime (since left to default) is seeing all of CPU cores on the node and starts that many native threads, with typically node cores > agent CPU limits, this increases likelihood of agent process having CPU throttled. See related known issue with Go https://github.com/golang/go/issues/33803

This PR configures GOMAXPROCS for sysdig container to be same as CPU requests to reduce CPU throttling. This way GOMAXPROCS stays in sync with chosen profile, no additional config var is introduced.

## Checklist

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -5,7 +5,7 @@ description: Sysdig Monitor and Secure agent
 type: application
 
 # currently matching sysdig 1.14.32
-version: 1.5.58
+version: 1.5.59
 
 appVersion: 12.10.1
 
--- a/charts/agent/templates/daemonset.yaml
+++ b/charts/agent/templates/daemonset.yaml
@@ -158,6 +158,11 @@ spec:
             readOnlyRootFilesystem: false
             allowPrivilegeEscalation: true
           env:
+            - name: GOMAXPROCS
+              valueFrom:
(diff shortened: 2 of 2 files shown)


aroberts87:
Hello @stevo-f3! Thank you for opening this PR. I've asked some of the agent developers about this and was told they are currently evaluating using `automaxprocs` (https://pkg.go.dev/go.uber.org/automaxprocs) and recommend we hold off on merging this PR until a decision is made. In the meantime, you can set the `GOMAXPROCS` value using the existing mechanism found here: https://github.com/sysdiglabs/charts/blob/42250b1e3388146b7dfb269f1fe857fae341e16f/charts/agent/values.yaml#L107 Any entries provided will get rendered into the daemonset here: https://github.com/sysdiglabs/charts/blob/42250b1e3388146b7dfb269f1fe857fae341e16f/charts/agent/templates/daemonset.yaml#L187-L190 Let me know if you need any assistance.

PR #884, opened 2023-01-11
feat(agent,node-analyzer,rapid-response): add Rancher-specific tolerations

ironashram:

Add Rancher-specific tolerations to various components to allow them to run in control-plane or etcd nodes by default

Successfully tested on a Rancher cluster

## Checklist

### DIFF
--- a/charts/node-analyzer/values.yaml
+++ b/charts/node-analyzer/values.yaml
@@ -106,8 +106,12 @@ nodeAnalyzer:
       key: node-role.kubernetes.io/master
     - effect: NoSchedule
       key: node-role.kubernetes.io/control-plane
+      operator: Equal
+      value: "true"
     - effect: NoExecute
       key: node-role.kubernetes.io/etcd
+      operator: Equal
+      value: "true"
 
   # Set nodeAnalyzer daemonset priorityClassName
   priorityClassName:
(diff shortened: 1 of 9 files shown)


charts/node-analyzer/values.yaml
```diff
@@ -106,6 +106,14 @@ nodeAnalyzer:
       key: node-role.kubernetes.io/master
     - effect: NoSchedule
       key: node-role.kubernetes.io/control-plane
+    - effect: NoSchedule
+      key: node-role.kubernetes.io/controlplane
```
lilx1ao, line 110: question: how does rancher know which key (controlplane or control-plane) to pick?

ironashram: kubeadm adds a taint `node-role.kubernetes.io/control-plane`, rancher instead uses `node-role.kubernetes.io/controlplane` by default, adding a toleration for both is the easiest way to allow the agent and other pods to be scheduled on those nodes

lilx1ao: I just read up on taints and tolerations and this starts to make sense. thank you!

lilx1ao (approved):

aroberts87 (approved):

PR #906, opened 2023-01-27
feat(registry-scanner): Add experimental cache PVC

tembleking:

Adds an experimental cache PVC

⚠️ currently, the usage of this cache is not threat-safe, so we cannot use it in the per-job-scanning.
waiting to resolve it before merging


wideawakening (approved):
let's try it 💪🏽

PR #911, opened 2023-02-01
feat(agent,sysdig-deploy): Implement global plan settings

aroberts87:
Introduce a new 'Sysdig Aware' subchart enable/disable feature that leverages global.sysdig.monitor and global.sysdig.secure to toggle:
1) Which subcharts of sysdig-deploy get rendered based on Sysdig deployment type
2) Which Agent settings are automatically set based on the type of deployment configured by the user.

This change also adds logic to enable secure_light mode in the Agent when specified.

The implementation here allows for the legacy <chart>.enabled values to override the new plan settings in an attempt at being backwards compatible.

## Checklist

### DIFF
--- a/charts/agent/README.md
+++ b/charts/agent/README.md
@@ -277,19 +277,6 @@ $ helm install --namespace sysdig-agent sysdig-agent -f values.yaml sysdig/agent
 You can read more details about this
 in [Kubernetes Documentation](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/).
 
-## Secure Light Mode
-[Secure Light](https://docs.sysdig.com/en/docs/installation/sysdig-agent/agent-configuration/configure-agent-modes/#secure-light) mode offers reduced Agent resource consumption by enabling only a curated subset of Sysdig Secure features. Those features are:
- - Runtime Policies
- - Activity Audit
- - Captures
-
-Sysdig Agent version 12.10 or above is needed for Secure Light mode. 
-
-To deploy the Sysdig Agent in Secure Light mode, run:
-```bash
-$ helm install --namespace sysdig-agent sysdig-agent --set sysdig.settings.features.mode=secure_light sysdig/agent
-```
-
 ## Modifying Sysdig agent configuration
 
 The Sysdig agent uses a file called `dragent.yaml` to store the configuration.
--- a/charts/agent/README.md
+++ b/charts/agent/README.md
@@ -277,6 +277,19 @@ $ helm install --namespace sysdig-agent sysdig-agent -f values.yaml sysdig/agent
 You can read more details about this
 in [Kubernetes Documentation](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/).
 
+## Secure Light Mode
+[Secure Light](https://docs.sysdig.com/en/docs/installation/sysdig-agent/agent-configuration/configure-agent-modes/#secure-light) mode offers reduced Agent resource consumption by enabling only a curated subset of Sysdig Secure features. Those features are:
+ - Runtime Policies
+ - Activity Audit
+ - Captures
+
+Sysdig Agent version 12.10 or above is needed for Secure Light mode. 
+
+To deploy the Sysdig Agent in Secure Light mode, run:
+```bash
+$ helm install --namespace sysdig-agent sysdig-agent --set sysdig.settings.features.mode=secure_light sysdig/agent
+```
+
 ## Modifying Sysdig agent configuration
 
 The Sysdig agent uses a file called `dragent.yaml` to store the configuration.
--- a/charts/agent/templates/_helpers.tpl
+++ b/charts/agent/templates/_helpers.tpl
@@ -335,7 +335,7 @@ and set the agent chart parameters accordingly
     {{- if and (and (hasKey .Values.global.sysdig "secure")
                     (not .Values.global.sysdig.secure))
                     .Values.secureFeatProvided }}
-        {{ fail "Cannot unset global.sysdig.secure when agent.sysdig.settings.feature.mode is secure or secure_light" }}
+        {{ fail "Set global.sysdig.secure=true when specifying agent.sysdig.settings.feature.mode=secure or agent.sysdig.settings.feature.mode=secure_light" }}
     {{- end }}
     {{- if and (and (hasKey .Values.global.sysdig "monitor")
                     .Values.global.sysdig.monitor)
--- a/charts/agent/templates/_helpers.tpl
+++ b/charts/agent/templates/_helpers.tpl
@@ -320,6 +320,58 @@ Use global sysdig tags for agent
 {{- end -}}
 
 {{/*
+Determine the plan settings (monitor/secure) set in the sysdig-deploy chart
+and set the agent chart parameters accordingly
+*/}}
+{{- define "agent.planSettings" -}}
+    {{/* Determine if secure or secure_light mode is being requested in Agent settings and store state */}}
+    {{- if hasKey .Values.sysdig.settings "feature" }}
+        {{- $_ := set .Values "secureLight" (eq .Values.sysdig.settings.feature.mode "secure_light") }}
+        {{- $_ := set .Values "secureFeatProvided" (or .Values.secureLight
+                                                       (eq .Values.sysdig.settings.feature.mode "secure")) }}
+    {{- end }}
+
+    {{/* Basic plan sanity checks */}}
+    {{- if and (and (hasKey .Values.global.sysdig "secure")
+                    (not .Values.global.sysdig.secure))
+                    .Values.secureFeatProvided }}
+        {{ fail "Cannot unset global.sysdig.secure when agent.sysdig.settings.feature.mode is secure or secure_light" }}
+    {{- end }}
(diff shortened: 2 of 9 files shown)


balinaveen (commented):
What about changes to Readme? I don't see the new global plan settings documented. 
It's not clear how secure_light should be enabled. It there a global plan setting that maps to this mode or is this an agent sub chart setting only?

aroberts87:
> What about changes to Readme? I don't see the new global plan settings documented. It's not clear how secure_light should be enabled. It there a global plan setting that maps to this mode or is this an agent sub chart setting only?

You beat me too it, @balinaveen. I didn't realize I was missing the README changes until I opened the PR. I'm making those changes now, but wanted to keep the PR open so people could take a look at what's there.

charts/sysdig-deploy/README.md
```diff
@@ -203,6 +205,42 @@ The following table lists the configurable parameters of the sysdig-deploy chart
...
+ - Activity Audit
+ - Captures
+
+Sysdig Agent version 12.10 or above is needed for Secure Light mode. 
+
+To deploy the Sysdig Agent in Secure Light mode, run:
+```bash
+$ helm install --namespace sysdig --set agent.sysdig.settings.features.mode=secure_light sysdig sysdig/sysdig-deploy
+```
+
```
lilx1ao: nice read 👍

charts/agent/README.md
```diff
@@ -277,6 +277,19 @@ $ helm install --namespace sysdig-agent sysdig-agent -f values.yaml sysdig/agent
...
+[Secure Light](https://docs.sysdig.com/en/docs/installation/sysdig-agent/agent-configuration/configure-agent-modes/#secure-light) mode offers reduced Agent resource consumption by enabling only a curated subset of Sysdig Secure features. Those features are:
+ - Runtime Policies
+ - Activity Audit
+ - Captures
+
+Sysdig Agent version 12.10 or above is needed for Secure Light mode. 
+
+To deploy the Sysdig Agent in Secure Light mode, run:
+```bash
+$ helm install --namespace sysdig-agent sysdig-agent --set sysdig.settings.features.mode=secure_light sysdig/agent
```
lilx1ao: why did you pick underscore over hyphen tho?

charts/agent/templates/_helpers.tpl
```diff
@@ -319,6 +319,58 @@ Use global sysdig tags for agent
...
+        {{- $_ := set .Values "secureLight" (eq .Values.sysdig.settings.feature.mode "secure_light") }}
+        {{- $_ := set .Values "secureFeatProvided" (or .Values.secureLight
+                                                       (eq .Values.sysdig.settings.feature.mode "secure")) }}
+    {{- end }}
+
+    {{/* Basic plan sanity checks */}}
+    {{- if and (and (hasKey .Values.global.sysdig "secure")
+                    (not .Values.global.sysdig.secure))
+                    .Values.secureFeatProvided }}
+        {{ fail "Cannot unset global.sysdig.secure when agent.sysdig.settings.feature.mode is secure or secure_light" }}
```
lilx1ao: I would add a suggestion "set global.sysdig.secure to be enabled" etc

charts/agent/README.md
aroberts87: Not my choice. 🙂 Anything under `sysdig.settings` is directly written into the ConfigMap and the agent itself uses
```
features:
  mode: secure_light
```
https://docs.sysdig.com/en/docs/installation/sysdig-agent/agent-configuration/configure-agent-modes/#secure-light

lilx1ao: I see!

charts/sysdig-deploy/Chart.yaml
```diff
@@ -21,28 +21,28 @@ dependencies:
     repository: file://../admission-controller
     version: ~0.7.23
     alias: admissionController
-    condition: admissionController.enabled
+    condition: admissionController.enabled,global.sysdig.plan.secure
```
lilx1ao: what's .plan?

charts/agent/templates/_helpers.tpl
```diff
@@ -319,6 +319,58 @@ Use global sysdig tags for agent
...
+For monitor.enabled=false, disable all Monitor specific extras (like app checks, prometheus, etc.)
+*/}}
+{{- define "agent.monitorFeatures" }}
+    {{- if not .Values.monitor.enabled }}
+        {{- $monitorBlock := dict "app_checks_enabled" false }}
+        {{- range $monitorFeature := (list
+            "jmx"
+            "prometheus"
+            "statsd") }}
+            {{- $_ := set $monitorBlock $monitorFeature (dict "enabled" false) }}
```
lilx1ao, line 368: I can guess what this block is trying to accomplish but my brain isn't allowing me to look too close today. as long as it's tested we cool

aroberts87: No worries, it's not trivial. It's assembling a dictionary that then gets rendered as YAML.
```yaml
app_checks_enabled: false
jmx:
  enabled: false
prometheus:
  enabled: false
statsd:
 enabled: false
```
I tested it regularly during development by manually inspecting the generated manifests, and there is [this check](https://github.com/sysdiglabs/charts/pull/911/files/5472a66391c3c1f31e15f193a0a0176e1bc7ffd3#diff-e33b13b6f8de45bf1e25d748554be6312f747666ae850b66c88d5a71c0638fc8R90-R93) for `app_checks_enabled: false` when secure only mode, but I can extend it to capture the whole block.

charts/sysdig-deploy/Chart.yaml
aroberts87: Good catch. 😬

charts/sysdig-deploy/README.md
```diff
@@ -203,6 +205,29 @@ The following table lists the configurable parameters of the sysdig-deploy chart
 | `rapidResponse`                         | Config specific to [Sysdig Rapid Response](#rapid-response)                                                             | `{}`      |
 | `rapidResponse.enabled`                 | Enable Rapid Response component in this chart                                                                           | `""`      |
 
+## Sysdig Plan Settings
```
balinaveen, line 208: The onboarding page in the UI shows this under "helm"

```
kubectl create ns sysdig-agent
helm repo add sysdig https://charts.sysdig.com
helm repo update
helm install sysdig-agent --namespace sysdig-agent \
    --set global.sysdig.accessKey=e64a1780-5bbd-4083-8ce1-418c2bb0a356 \
    --set global.sysdig.region=custom \
    --set nodeAnalyzer.secure.vulnerabilityManagement.newEngineOnly=true \
    --set global.kspm.deploy=true \
    --set nodeAnalyzer.nodeAnalyzer.benchmarkRunner.deploy=false \
    --set global.clusterConfig.name=<CLUSTER_NAME> \
    sysdig/sysdig-deploy 
```
How will that change with the introduction of the new global plan settings? For example, are the nodeAnalyzer settings still required?

lilx1ao (approved):

charts/sysdig-deploy/README.md
aroberts87: With these changes as they are the instructions could become slightly more complicated as secure customers can now optionally set `global.sysdig.secure=true` in the `helm install ...` command. The reason for this is because the other settings, particularly `nodeAnalyzer.secure.vulerabilityManagement.newEngineOnly=true` and `nodeAnalyzer.nodeAnalyzer.benchmarkRunner.deploy=false` are manipulating how the subcharts work.

One change I can do is include the the KSPM collector as a part of `global.sysdig.secure=true`. There was actually a whole discussion on this topic yesterday (disabling bencmark v2 and enabling kspm-collector when `global.sysdig.secure` is set). I was planning on doing that change in a follow up PR, but it wouldn't be too hard to put into this one. @balinaveen , what are your thoughts?

aroberts87: >One change I can do is include the the KSPM collector as a part of global.sysdig.secure=true. There was actually a whole discussion on this topic yesterday (disabling bencmark v2 and enabling kspm-collector when global.sysdig.secure is set). I was planning on doing that change in a follow up PR, but it wouldn't be too hard to put into this one. @balinaveen , what are your thoughts?

On second thought, I would not like to do this as a part of this PR. I would like to engage the KSPM folks for this change so that they are aware.

balinaveen: Yeah, let's do the ksmp changes later. If the onboarding UI as it exists today will still work, then that's OK. I was hoping things would be simpler, but that can wait. Let's postpone the update to the UI until all changes needed to simplify things are in the helm chart.

claffin: We should implement this with KSPM collector and KSPM analyzer as a part of global.sysdig.secure=true.

KSPM is now GA and the benchmark runner is now deprecated. We shouldn't give these plans to customers without this otherwise they'll be missing features, especially new customers during a PoV/onboarding.

claffin (changes requested):

aroberts87:
Based on some discussions about this PR we have opted to close this single PR and instead do the following:
1) Close this PR and break out the agent chart specific improvements and get those in a separate PR.
2) Wait for the benchmark runner --> kspm migration to happen
3) Open a new PR for these plan setting improvements after the aforementioned migration has occurred.

PR #907, opened 2023-01-30
fix(node-analyzer,kspm-collector): Fix security context permission for KSPM components

dark-vex:
## What this PR does / why we need it:

When deploying sysdig-deploy chart with the flag `--set global.kspm.deploy=true` (which is the default now), `node-analyzer` (due to kspm-analyzer container) and `kspm-collector` pods are not getting deployed on OpenShift due to OCP SCC permission.

For `node-analyzer` kspm-analyzer container we are exposing container and host port 12000 for the readiness/liveness check which is not required because kubernetes/ocp is capable to run this health-checks without exposing the container port. I have removed the `containerPort` and adjusted the readiness/liveness check for kspm-analyzer container.

For `kspm-collector` same thing, in this case we don't have the manifest that request the OCP SCC permission, so in addition to the containerPort change, I have also added `securitycontextconstraint.yaml` manifest

### Prior the change
```
 ➜ oc get ds,deploy -n sysdig-agent
NAME                                        DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
daemonset.apps/sysdig-agent                 6         6         4       6            4           <none>          2m14s
daemonset.apps/sysdig-agent-node-analyzer   0         0         0       0            0           <none>          2m14s

NAME                                         READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/sysdig-agent-kspmcollector   0/1     0            0           2m14s
```

### After the change
```
 ➜ oc get po -n sysdig-agent
NAME                                          READY   STATUS    RESTARTS   AGE
sysdig-agent-6kzwd                            1/1     Running   0          11m
sysdig-agent-8nvvm                            1/1     Running   0          11m
sysdig-agent-dq5g7                            1/1     Running   0          11m
sysdig-agent-kspmcollector-857fcdf96b-z4g8q   1/1     Running   0          3m13s
sysdig-agent-mhc9t                            1/1     Running   0          11m
sysdig-agent-node-analyzer-58mfg              3/3     Running   0          3m14s
sysdig-agent-node-analyzer-6rkql              3/3     Running   0          3m14s
sysdig-agent-node-analyzer-bdtdd              3/3     Running   0          3m14s
sysdig-agent-node-analyzer-bh945              3/3     Running   0          3m14s
sysdig-agent-node-analyzer-pcncf              3/3     Running   0          3m14s
sysdig-agent-node-analyzer-vbhv7              3/3     Running   0          3m14s
[...]

PR #916, opened 2023-02-07
feat(node-analyzer): bumped RuntimeScanner and HostScanner to latest versions

ProvoK:
Extended Changelog:
* HostScanner, bumped to 0.3.6: 
  * Security fix for CVE-2022-41721 
  * Fix AcceptRisk evaluation wrong in some edge cases
* RuntimeScanner, bumped to 1.4.6: 
  * Fix AcceptRisk evaluation wrong in some edge cases

NOTE: IIRC changelog changes are not needed if we provide the correct commit body payload. Please lmk if it's not the case @sysdiglabs/team-tools-agent


charts/sysdig-deploy/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sysdig-deploy
 description: A chart with various Sysdig components for Kubernetes
 type: application
-version: 1.5.57
```
aroberts87: Hey @ProvoK ! You no longer need to manually increment the  `sysdig-deploy` chart version or the dependency versions, we have a GitHub action that does that automatically now if a dependent subchart is modified (https://github.com/sysdiglabs/charts/blob/master/.github/workflows/update-sysdig-deploy-chart.yaml). Can you please remove the edits to `charts/sysdig-deploy/Chart.yaml`?

aroberts87 (changes requested):

charts/sysdig-deploy/Chart.yaml
ProvoK: sure! very happy to do that 😄

ProvoK: just FYI: `node-analyzer` was at version `x.x.28` and `sysdig-deploy` was referencing `x.x.27`, so maybe something it's not working as intended, you may want to double check.

ProvoK: rollbacked changes with [8b314fc](https://github.com/sysdiglabs/charts/pull/916/commits/8b314fc5409e92426d13848d92131842d49613ee)

aroberts87: > just FYI: `node-analyzer` was at version `x.x.28` and `sysdig-deploy` was referencing `x.x.27`, so maybe something it's not working as intended, you may want to double check.

I'm going to look into this now. This shouldn't happen. 🤔

aroberts87 (approved):

charts/sysdig-deploy/Chart.yaml
aroberts87: @ProvoK - `sysdig-deploy` is referencing `x.x.28` now (https://github.com/sysdiglabs/charts/blob/master/charts/sysdig-deploy/Chart.yaml#L34), but I think I now understand where the confusion is coming from. The `sysdig-deploy` chart is not bumped as a part of the merge containing the subchart changes, but rather a new PR is opened automatically with the version bumps. In this case that autogenerated PR is here: https://github.com/sysdiglabs/charts/pull/914. Those autogenerated PRs are approved by someone on @sysdiglabs/team-tools-agent. The version of `master` you were looking at may have been from a time between when the `node-analyzer` chart was bumped to `x.x.28` (https://github.com/sysdiglabs/charts/pull/907) and when I approved and merged the `sysdig-deploy` version bumps. Does that make sense?

PR #917, opened 2023-02-07
feat(agent): Add new monitor.enabled flag to agent chart

aroberts87:
The new monitor.enabled flag acts as an analogue to the existing secure.enabled
and can be used for enabling or forcibly disabling Sysdig Monitor components
in the Sysdig Agent.

## Checklist


charts/agent/templates/_helpers.tpl
```diff
@@ -333,26 +376,33 @@ secure.enabled=false we need to explicitly put those config entires in the
 agent config to prevent a backend push from enabling them after installation.
 */}}
 {{- define "agent.secureFeatures" }}
-    {{- $secureBlockConfig := dict "security" (dict
-        "enabled" .Values.secure.enabled
-        "k8s_audit_server_enabled" .Values.auditLog.enabled) }}
+    {{- if .Values.auditLog.enabled }}
```
deepskyblue86: Please put a comment to describe what we're doing here, as opposite as what we're doing in the next if auditLog.enabled

charts/agent/templates/_helpers.tpl
```diff
@@ -319,6 +319,49 @@ Use global sysdig tags for agent
...
 
+{{/*
+Determine the plan settings (monitor/secure) set in the sysdig-deploy chart
+and set the agent chart parameters accordingly
+*/}}
+{{- define "agent.planSettings" -}}
+    {{/* Determine if secure or secure_light mode is being requested in Agent settings and store state */}}
+    {{- if hasKey .Values.sysdig.settings "feature" }}
+        {{- $_ := set .Values "secureLight" (eq .Values.sysdig.settings.feature.mode "secure_light") }}
+        {{- $_ := set .Values "secureFeatProvided" (or .Values.secureLight
```
deepskyblue86: Why are we storing those values? Can't we just have them as variables since they're used only here?

charts/agent/templates/_helpers.tpl
```diff
@@ -333,26 +376,33 @@ secure.enabled=false we need to explicitly put those config entires in the
...
 */}}
 {{- define "agent.secureFeatures" }}
-    {{- $secureBlockConfig := dict "security" (dict
-        "enabled" .Values.secure.enabled
-        "k8s_audit_server_enabled" .Values.auditLog.enabled) }}
+    {{- if .Values.auditLog.enabled }}
+        {{- if or (and (hasKey .Values.global.sysdig "secure")
+                       (not .Values.global.sysdig.secure))
+                   .Values.secureLight }}
+            {{- $_ := set .Values.auditLog "enabled" false }}
```
deepskyblue86: Is there a way to have a (debug?) message to the user to notify them that we're overruling what they asked?

deepskyblue86: I'm not sure this should be in the scope of this PR

charts/agent/templates/_helpers.tpl
```diff
@@ -333,26 +376,33 @@ secure.enabled=false we need to explicitly put those config entires in the
...
-                 "k8s_audit_server_url" .Values.auditLog.auditServerUrl
-                 "k8s_audit_server_port" .Values.auditLog.auditServerPort) }}
-            {{- $_ := set $secureBlockConfig.security $key $val }}
+             "k8s_audit_server_url" .Values.auditLog.auditServerUrl
+             "k8s_audit_server_port" .Values.auditLog.auditServerPort) }}
+        {{- $_ := set $secureConfig.security $key $val }}
         {{- end }}
     {{- end }}
-    {{- if not .Values.secure.enabled }}
+    {{- if or (not .Values.secure.enabled)
```
deepskyblue86: I'm a bit puzzled about checking this vs checking global.sysdig, could you please elaborate it a bit?

charts/agent/README.md
```diff
@@ -118,6 +118,7 @@ The following table lists the configurable parameters of the Sysdig chart and th
 | `sysdig.settings`                              | Additional settings, directly included in the agent's configuration file `dragent.yaml`                                                                 | `{}`                                                        |
 | `logPriority`                                  | Use to directly set both Agent console and file logging priorities. Possible values: `"info"`, `"debug"`. Mutually exclusive with `sysdig.settings.log` | ` `                                                         |
 | `secure.enabled`                               | Enable Sysdig Secure                                                                                                                                    | `true`                                                      |
+| `monitor.enabled`                              | Enable Sysdig Monitor                                                                                                                                   | `true`                                                      |
```
deepskyblue86, line 121: Aren't we supposed to have a changelog line to document it too?

charts/agent/tests/secure_enable_test.yaml
```diff
@@ -75,3 +75,50 @@ tests:
...
+        enabled: false
+      sysdig:
+        settings:
+          feature:
+            mode: secure_light
+    asserts:
+      - failedTemplate:
+          errorMessage: 'Set agent.secure.enabled=true when specifying agent.sysdig.settings.feature.mode is secure or secure_light'
+
+  - it: Test monitor.enabled when specifying secure_light mode
```
deepskyblue86:
```suggestion
  - it: Test monitor.enabled when specifying secure mode
```

charts/agent/tests/monitor_enable_test.yaml
```diff
@@ -0,0 +1,59 @@
...
+        enabled: true
+      sysdig:
+        settings:
+          feature:
+            mode: secure_light
+    asserts:
+      - failedTemplate:
+          errorMessage: 'Cannot set agent.monitor.enabled=true when agent.sysdig.settings.feature.mode is secure or secure_light'
+
+  - it: Test monitor.enabled when specifying secure_light mode
```
deepskyblue86:
```suggestion
  - it: Test monitor.enabled when specifying secure mode
```

deepskyblue86 (commented):
There are some points that are not really clear to me.
Even if overall it looks fine, I don't really feel entitled to cast a vote.
I hope my comments would be somehow useful at least.

charts/agent/README.md
aroberts87: The changelog entires are now automatically generated via a GitHub action defined in [this](https://github.com/sysdiglabs/charts/tree/master/.chglog) directory.

charts/agent/templates/_helpers.tpl
aroberts87: They need to be stored because they are being used in the `agent.secureFeatures` helper function. I am doing the setting here so I don't need to do it twice, but could move it around if you feel it's less obscure.

charts/agent/templates/_helpers.tpl
aroberts87: > I'm not sure this should be in the scope of this PR

That's a fair point. This branch started out with much grander visions about a global secure/monitor flags for enabling and disabling, hence the checks in `global.sysdig`, but the larger vision was postponed. This is an artifact from that time and could be removed, but it would result in a slight smaller config being generated if the user either disables secure mode or runs in secure light since the subsequent `if` block will not render if `auditLog.enabled` is false.

tl;dr - I agree this is out of the scope of this PR and I'll remove it.

charts/agent/templates/_helpers.tpl
aroberts87: I'll address this comment by taking out the subsequent block as I agree with your statement that it's outside the scope of this PR.

charts/agent/templates/_helpers.tpl
aroberts87: I addressed this in my previous comment, but the use of `global.sysdig` vs `Values.secure` was from a previous iteration of this branch where there was an attempt at putting higher logic in the charts to dictate which ones rendered vs didn't based on user type. The decision was made to trim down the scope of the branch, but the previous block had erroneously not been updated from `global.sysdig` back to the agent chart specific `Values.secure`.

charts/agent/templates/_helpers.tpl
```diff
@@ -350,29 +391,39 @@ secure.enabled=false we need to explicitly put those config entires in the
...
+    }}
         {{- range $secureFeature := (list
             "commandlines_capture"
             "drift_killer"
             "falcobaseline"
             "memdump"
             "network_topology"
             "secure_audit_streams") }}
-            {{- $_ := set $secureBlockConfig $secureFeature (dict "enabled" false) }}
+            {{- $secureConfig := set $secureConfig $secureFeature (dict "enabled" false) }}
```
mavimo: @mavimo double check if is correct or need to be replaced with:
```
{{- $_ := set $secureConfig $secureFeature (dict "enabled" false) }}
```

mavimo:
Test matrix

| Secure | Monitoring | Status | 
|--------|------------|--------|
| true | true | ✅ |
| true | false | ✅ |
| false | true | ✅ |
| false | false | ✅ |

mavimo (approved):

PR #945, opened 2023-02-21
feat(agent): Create Delegated Agent installation mechanism

aroberts87:
Add the ability to install the agent in a specialized delegated manner where only a small portion of the overall nodes in the cluster communicate with the Kubernetes API server. This is accomplished by installing a modified Agent DaemonSet with an additional Agent Deployment. The Agents in the Deployment are the ones which communicate with the Kubernetes API server, but don't do much, if anything, else.

## Checklist


charts/agent/templates/configmap.yaml
```diff
@@ -1,67 +1,32 @@
...
-    k8s_coldstart:
-      enabled: true
-      enforce_leader_election: true
-      namespace: {{ include "agent.namespace" . }}
-{{- end }}
-{{- if .Values.global.sysdig.tags }}
+    {{- include "agent.clusterName" . | indent 4 }}
+    {{- if .Values.delegatedAgentDeployment.enabled }}
+    cointerface_enabled: false
+    {{- end }}
```
irozzo-1A: nit: it could be useful to deploy the agent with cointerface enabled even with the agent deployment. It would help to have:
1. More uniform resource consumption for the agents (no delegated)
2. No need for running delegation process
If this is not the primary goal we can tackle this later

charts/agent/templates/configmap.yaml
```diff
@@ -1,67 +1,32 @@
-{{- if .Values.daemonset.deploy }}
+{{- if or .Values.daemonset.deploy .Values.delegatedAgentDeployment.enabled }}
```
irozzo-1A, line 1: Shouldn't we provide the possibility to deploy the deployment without the daemonset?

charts/agent/templates/configmap.yaml
```diff
@@ -1,67 +1,32 @@
...
-{{- if .Values.leaderelection.enable }}
-    k8s_delegation_election: true
-    k8s_coldstart:
-      enabled: true
-      enforce_leader_election: true
-      namespace: {{ include "agent.namespace" . }}
-{{- end }}
-{{- if .Values.global.sysdig.tags }}
+    {{- include "agent.clusterName" . | indent 4 }}
+    {{- if .Values.delegatedAgentDeployment.enabled }}
```
irozzo-1A: This is causing trailing spaces:
```
    new_k8s: true    
    k8s_cluster_name: agent-deployment
    cointerface_enabled: false    
```
Maybe we should do something like:
```suggestion
    {{- include "agent.clusterName" . | trim | nindent 4 }}
    {{- if .Values.delegatedAgentDeployment.enabled }}
```

charts/agent/templates/configmap.yaml
```diff
@@ -1,67 +1,32 @@
...
+    cointerface_enabled: false
+    {{- end }}
+    {{- include "agent.connectionSettings" . | indent 4 }}
```
irozzo-1A: Same here:
```suggestion
    {{- include "agent.connectionSettings" . | trim | nindent 4 }}
```

charts/agent/templates/configmap-deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+{{ include "agent.labels" . | indent 4 }}
+data:
+  dragent.yaml: |
+    new_k8s: true
+    {{- include "agent.clusterName" . | indent 4 }}
+    {{- include "agent.connectionSettings" . | indent 4 }}
+    feature:
+      mode: monitor_light
+    cointerface_enabled: true
+    k8s_delegated_nodes: -1
```
irozzo-1A: We could think about making this configurable

charts/agent/values.yaml
```diff
@@ -275,6 +275,41 @@ tolerations:
...
+    arch:
+      - amd64
+      - arm64
+      - s390x
+    env: {}
+    os:
+      - linux
+    labels: {}
+    progressDeadlineSeconds: 600
+    replicas: 2
```
irozzo-1A: I would rather use 1 as the default, to avoid having two pods to send the global kubernetes state

charts/agent/templates/daemonset.yaml
```diff
@@ -1,12 +1,12 @@
-{{- if .Values.daemonset.deploy }}
+{{- if or .Values.daemonset.deploy .Values.delegatedAgentDeployment }}
```
irozzo-1A, line 1: Same here, not sure we should couple the deployment with agent daemonset

irozzo-1A (commented):
Overall LGTM, I think that the whitespaces handling in the agent config should be double-checked, as it's breaking the way it's rendered:

Got:
```
$ k get cm sysdig-agent-deployment -oyaml
apiVersion: v1
data:
  dragent.yaml: "new_k8s: true    \nk8s_cluster_name: agent-deployment    \ncollector:
    collector-staging.sysdigcloud.com\ncollector_port: 6443\nfeature:\n  mode: monitor_light\ncointerface_enabled:
    true\nk8s_delegated_nodes: -1\ntop_processes_in_sample: 0\ncontainer_engines:\n
    \ bpm: false\n  cri: false\n  docker: false\n  libvirt_lxc: false\n  lxc: false\n
    \ mesos: false\n  podman: false\n  rkt: false\napp_checks_enabled: false\njmx:\n
    \ enabled: false\nprometheus:\n  enabled: false\nstatsd:\n  enabled: false\nk8s_coldstart:\n
    \ max_parallel_cold_starts: 1\n\ncommandlines_capture:\n  enabled: false\ndrift_killer:\n
    \ enabled: false\nfalcobaseline:\n  enabled: false\nmemdump:\n  enabled: false\nnetwork_topology:\n
    \ enabled: false\nsecure_audit_streams:\n  enabled: false\nsecurity:\n  enabled:
    false\n  k8s_audit_server_enabled: true\n  k8s_audit_server_port: 7765\n  k8s_audit_server_url:
    0.0.0.0\n\n\napp_checks_enabled: false\nimds_version: 2\nlog:\n  file_priority:
    debug\n  max_size: 100\n  rotate: 25\nmetricsfile:\n  json: true\n  location:
    metrics\n"
```

Expected:
```
$ k get cm sysdig-agent -oyaml
apiVersion: v1
data:
  dragent.yaml: |
    new_k8s: true
    k8s_cluster_name: agent-deployment
    cointerface_enabled: false
    collector: collector-staging.sysdigcloud.com
    collector_port: 6443
    security:
      enabled: true
      k8s_audit_server_enabled: true
      k8s_audit_server_port: 7765
      k8s_audit_server_url: 0.0.0.0
    k8s_coldstart:
      max_parallel_cold_starts: 1


    app_checks_enabled: false
    imds_version: 2
    log:
      file_priority: debug
      max_size: 100
      rotate: 25
    metricsfile:
      json: true
      location: metrics
```

charts/agent/templates/configmap.yaml
aroberts87: That's an interesting point, but what would that provide the customer? Just k8s metadata?

charts/agent/templates/configmap.yaml
aroberts87: Hey @irozzo-1A , could you please elaborate on this? My understanding was that the whole purpose of having the deployment was to migrate cointerface functionality there. What would be purpose of having two cointerface processes running on the same node?

irozzo-1A: sure @aroberts87 , the main purpose of this is to have one only Pod talking to Apiserver. This would be another use-case for average-size clusters where the Apiserver load is not a concern.
As the delegated agent consumes more resources than the others (especially memory), this would provide more uniformity over the agents controlled by the ds (it will be easier to set requests/limits). Another benefit would be not having to run the delegation election process (should be stable in the last agent versions, but we had some bugs in the past).

charts/agent/templates/configmap.yaml
irozzo-1A: Yes, for instance, secure features that only require an agent to send global kubernetes state to BE, could benefit from this.
I'm going to double-check if there is some 😄 .

alblasco (approved):

charts/agent/templates/configmap.yaml
mavimo: @irozzo-1A considering the implication (capabilities that we are going to lose) and the fact that we discourage the usage of this feature, I'll suggest to keep it only ad the solution in case of large clusters, WDYT?

irozzo-1A: @mavimo I think this could turn useful for average-sized clusters, but don't consider this as a blocking point for this PR. We can address it later in case of need.

mavimo (approved):

PR #946, opened 2023-02-22
feat(admission-controller,agent,common,kspm-collector,node-analyzer,rapid-response,sysdig-deploy): Improve region resolution in all charts

aroberts87:
* Refactor Collector and API endpoint determinations from the individual child charts of sysdig-deploy and move them to a new library chart called 'common'. All charts can now list that library chart as a dependency and share the common logic to resolve the various Collector and API endpoints from a given region code.
* Added IBM regions

## Checklist

### DIFF
--- a/.github/workflows/helm-unit-test.yaml
+++ b/.github/workflows/helm-unit-test.yaml
@@ -19,8 +19,14 @@ jobs:
       - name: Set up helm unit test plugin
         run: helm plugin install https://github.com/helm-unittest/helm-unittest --version=0.3.0
 
-      - name: Bundle sysdig-deploy dependencies
-        run: helm dependency build ./charts/sysdig-deploy
+      - name: Bundle chart dependencies
+        run: |
+          helm dependency build ./charts/admission-controller
+          helm dependency build ./charts/agent
+          helm dependency build ./charts/kspm-collector
+          helm dependency build ./charts/node-analyzer
+          helm dependency build ./charts/rapid-response
+          helm dependency build ./charts/sysdig-deploy
 
       - name: Test admission-controller
         run: helm unittest --strict ./charts/admission-controller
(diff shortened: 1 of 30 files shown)


charts/common/templates/_regions.tpl
```diff
@@ -0,0 +1,200 @@
+{{- define "sysdig.regions" -}}
```
AlbertoBarba, line 1: I can't find anywhere in the IBM documentation where these endpoints are defined:
```
- <region>.sysdig-secure.cloud.ibm.com
- ingest.<region>.sysdig-secure.cloud.ibm.com
- ingest.private.<region>.sysdig-secure.cloud.ibm.com
- private.<region>.sysdig-secure.cloud.ibm.com
```

.github/workflows/helm-unit-test.yaml
```diff
@@ -19,8 +19,14 @@ jobs:
       - name: Set up helm unit test plugin
         run: helm plugin install https://github.com/helm-unittest/helm-unittest --version=0.3.0
 
-      - name: Bundle sysdig-deploy dependencies
-        run: helm dependency build ./charts/sysdig-deploy
+      - name: Bundle chart dependencies
+        run: |
+          helm dependency build ./charts/admission-controller
```
AlbertoBarba: Maybe we can add a new target in the `Makefile` that executes it on every chart?
```Makefile
deps:
	find ./charts -name "Chart.yaml" | \
		xargs -L1 dirname | \
		xargs -I% sh -c \
			"helm dependency build %"
```

charts/sysdig-deploy/Chart.yaml
```diff
@@ -14,19 +14,23 @@ dependencies:
...
     alias: admissionController
     condition: admissionController.enabled
   - name: agent
     # repository: https://charts.sysdig.com
     repository: file://../agent
-    version: ~1.9.2
+    version: ~1.9.3
     alias: agent
     condition: agent.enabled
+  - name: common
```
AlbertoBarba, line 26: Do we really need to add `common` also in `sysdig-deploy`?

AlbertoBarba (approved):

PR #959, opened 2023-02-27
ci(admission-controller,agent,kspm-collector,node-analyzer,rapid-response,sysdig-deploy): add trigger to helm chart validation tests upon PR

macuna-sd:

This PR creates a new github actions pipeline that will be triggered when changes are done to charts. This pipeline in turn triggers helm-chart-validation pipeline.

## Checklist


.github/workflows/chart-automation-tests.yaml
```diff
@@ -0,0 +1,31 @@
...
+on:
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'charts/sysdig-deploy/**'
+      - 'charts/admission-controller/**'
+      - 'charts/node-analyzer/**'
+      - 'charts/kspm-collector/**'
+      - 'charts/rapid-response/**'
```
aroberts87, line 12: You're missing the agent chart. 🙂

.github/workflows/chart-automation-tests.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  helm-charts-validation-test:
+    runs-on: tools-runner
```
aroberts87: Is this a custom runner that needs to be created before this is merged?

.github/workflows/chart-automation-tests.yaml
macuna-sd: sorry what do you mean by that?

.github/workflows/chart-automation-tests.yaml
macuna-sd: you may be right...

.github/workflows/chart-automation-tests.yaml
aroberts87: You need to add
```
- 'charts/agent/**'
```
🙂

.github/workflows/chart-automation-tests.yaml
aroberts87: Is there anything special on that runner? If there isn't anything there that is proprietary you could just use something like `ubuntu-latest` and use apt to install anything that you need during each run. 🤷🏻‍♂️

PR #961, opened 2023-02-28
feat(sysdig-agent,node-analyzer,kspm-collector,rapid-response,admission-controller): Global Custom CA Bundle Support

saltycr3w:

This PR adds Global Custom CA Bundles support inside the relevant components.

- `sysdig-agent` (Complete)
- `node-analyzer` (Complete)
- `kspm-collector` (Complete)
- `rapid-response` (Complete)
Rapid Response has local chart Custom CA Support already, this PR takes that into account and adds onto it.
- `admission-controller` (Complete)
admission-controller has local chart Custom CA Support for the scanner and webhook, this PR takes that into account and adds onto it.

## TODO

Add more documentation about usage.

## Checklist


charts/admission-controller/values.yaml
```diff
@@ -223,11 +256,24 @@ webhook:
       # <br/>And inbound connections to serve HttpRequests as Kubernetes Webhook.
       # <br/>A PEM-encoded x509 certificate authority.
       # <br/>If empty, a new CA will be autogenerated.
-      cert: ""
+      cert:
```
AlbertoBarba: @aroberts87 @mavimo 
Using the `webhook.ssl.ca.cert` value for the scope of this PR might actually cause a BC.
This is currently used in https://github.com/sysdiglabs/charts/blob/master/charts/admission-controller/templates/_helpers.tpl#L166 in order to generate a cert that can be later used in https://github.com/sysdiglabs/charts/blob/master/charts/admission-controller/templates/webhook/admissionregistration.yaml#L42
Since `global.ssl.ca.cert` and `<component>.ssl.ca.cert` actually contains more than one cert, it might actually make sense to rename it in `certs` (actually `rapid-response` and `registry-scanner` already use `certs` for this purpose).
WDYT?

mavimo: `certs` LGTM (also because it can contains multiple certs 😅 )
Are you considering to make it an array or just a string with multiple certs?

saltycr3w: @AlbertoBarba @mavimo @aroberts87 

The reason I didn't touch the local CA on these components was they were written before the global.  now, technically a user should never need more than 1 root CA to add, however, you can bundle them into 1 file per SSL RFC.  So this is why I opted to just do a string in the global section as they can concatenate as many as they want.

I couldn't change the local's on these because we don't know if any customers are using them to break their deployments.

Just my last 2c while I'm here.  :)

AlbertoBarba: @dthomson-sysdig exactly...as u said we are not going to change the local CA of these components. In order to keep it in place and avoid any possible conflict with the new config we are just going to differentiate the values names. Even if we keep it as a string I'm not 100% sure that https://github.com/sysdiglabs/charts/blob/master/charts/admission-controller/templates/_helpers.tpl#L166 will work properly if a customer specifies more than one CA. I guess we'll have to figure out how to solve it once we'll decide to make a major release and do a BC.

charts/rapid-response/values.yaml
```diff
@@ -6,6 +6,40 @@ global:
...
+      #   -----END CERTIFICATE-----
+      #   -----BEGIN CERTIFICATE-----
+      #   MIIDEzCCAfugAwIBAgIQKiv9U+KxPJzu1adXwC06RzANBgkqhkiG9w0BAQsFADAU
+      #   MRIwEAYDVQQDEwloYXJib3ItY2EwHhcNMjIwMjIzMDY1NjExWhcNMjMwMjIzMDY1
+      #   NjExWjAUMRIwEAYDVQQDEwloYXJib3ItY2EwggEiMA0GCSqGSIb3DQEBAQUAA4IB
+      #   MMNlTAQ9fvdNOTzZntye0PQYRTTS34D=
+      #   -----END CERTIFICATE-----
+
+      # Filename that is used when creating the secret.  Required if cert is provided.
+      keyName: root_ca_file.crt
```
mavimo, line 33: Remove the default here (the default is missing in all the other entries)

charts/rapid-response/values.yaml
```diff
@@ -6,6 +6,40 @@ global:
   image:
     pullSecrets: []
     pullPolicy: IfNotPresent
+  ssl:
+    ca:
+      # For outbound connections (secure backend, proxy,...)
+      # A PEM-encoded x509 certificate.  This can also be a bundle with multiple certificates.
+      # See below in the rapidResponse.ssl.ca.cert key for an example
+      cert:
```
mavimo: This should be `certs: []` to keep the same strcuture we have in the other components.

mavimo (changes requested):

charts/rapid-response/values.yaml
AlbertoBarba: Removing the default here will cause a BC since the current version of `rapid-response` already accept the `.Values.ssl.ca.certs` and automatically names it in https://github.com/sysdiglabs/charts/pull/961/files#diff-4dd3290c87066c244c1311d77bc14f6a87e30ceb1bdf42ca0f67239c738854e6L38

mavimo: Ok. Semi-related question: why we don't have the default file name in all the charts? 😅

AlbertoBarba: Cause almost everywhere else we didn't have this functionality yet...the name has been added here to not cause any issue to who's already using this on `rapid-response`.

mavimo (approved):

AlbertoBarba (approved):

PR #962, opened 2023-02-28
ci(registry-scanner): add integration test for aws single

wideawakening:
allow registry-scanner chart test for aws single-account


hayk99 (approved):
lgtm!

PR #968, opened 2023-03-01
feat(agent): Add poddisruptionbudget permissions to the Agent clusterrole

bawhetst:

We're adding the capability to send PodDisruptionBudget KSM metrics from the Agent. This change adds the necessary permissions to get, list, and watch the PodDisruptionBudget resources necessary for generating these metrics. This feature is planned for the major Agent release after 12.12.0.

## Checklist


aroberts87:
@bawhetst - Is the plan to release this at the same time as the next agent release? And a follow up, is this feature only relevant for users with specific monitor features enabled?

bawhetst:
> Is the plan to release this at the same time as the next agent release? 

I'd like to release it as soon as possible since our automation uses Helm to install the Agent. As of right now, the automated test that QA wrote to test this functionality is failing for the dev branch since the clusterrole doesn't allow the Agent to access `poddisruptionbudgets`.

> And a follow up, is this feature only relevant for users with specific monitor features enabled?

Enabling the functionality does require a config change but is not an Agent feature. It's controlled by the `k8s_extra_resources` config parameter. However, the cointerface feature must be enabled in order for any Kubernetes monitoring to take place.

aroberts87 (approved):

PR #970, opened 2023-03-02
refactor(registry-scanner): simplify filter creation in configmap

tembleking:
## Checklist


charts/registry-scanner/values.yaml
```diff
@@ -67,9 +67,9 @@ config:
     # List of regular expressions. Images matching any of these expressions are excluded when scanning.
     exclude: []
     # Exclude images with creation date older than specified number of days.
-    maxAgeDays:
+    maxAgeDays: 0
```
airadier, line 70: I would suggest commenting here and in the README.md that the value "0" means unlimited or no filter. Otherwise it might not be straightforward for the user.

charts/registry-scanner/templates/configmap.yaml
```diff
@@ -71,17 +71,7 @@ data:
...
-      include:
-        {{- toYaml .Values.config.filter.include | nindent 8 }}
-      {{- end }}
-      {{- if .Values.config.filter.exclude }}
-      exclude:
-        {{- toYaml .Values.config.filter.exclude | nindent 8 }}
-      {{- end }}
-      maxAgeDays: {{ default 0 .Values.config.filter.maxAgeDays }}
-      maxTagsPerRepository: {{ default 0 .Values.config.filter.maxTagsPerRepository }}
+    filter: {{- toYaml .Values.config.filter | nindent 6 }}
```
airadier, line 74: Nice catch! I wonder how the block above ended up being so convoluted 🤔

airadier (changes requested):

PR #975, opened 2023-03-04
feat(agent): Fix GKE Autopilot installation

aroberts87:
There are several items related to the Agent chart that do not work when installing into a GKE Autopilot cluster. This change does the following to address them:

- Set a 500Mi ephemeral storage request on the agent container so that the containers are guaranteed enough space to compile the BPF probe
- Either take a priority class name, or optionally create a new priority class, to ensure the Agent pods get scheduled on the initial set of small nodes provided by Autopilot
- Enforce the use of the regular (non-slim) agent container in GKE Autopilot 

## Checklist


saltycr3w (approved):
Looks good, and it's configurable, yay!

Did you you have to move the merge into the daemonset.yaml instead of inside the _helper for a particular reason?  Just curious.

dtaracks (approved):
So much better now - Thanks

aroberts87:
> Looks good, and it's configurable, yay!
> 
> Did you you have to move the merge into the daemonset.yaml instead of inside the _helper for a particular reason? Just curious.

@dthomson-sysdig - Are you talking about for the kmodule container? I left it in the daemonset instead of moving to the _helper file because those values were just being taken directly from the `values.yaml` and don't ever need to comply with the resource profiles, so I figured it would reduce the complexity to just handle the splicing locally. Does that make sense?

charts/agent/templates/daemonset.yaml
```diff
@@ -1,4 +1,7 @@
 {{- if .Values.daemonset.deploy }}
+{{- if and .Values.gke.autopilot .Values.slim.enabled }}
```
mavimo: use `{{- include "agent.gke.autopilot" . }}` to check for autopilot using the helper.

charts/agent/templates/daemonset.yaml
```diff
@@ -29,8 +32,8 @@ spec:
       {{- end }}
     spec:
       serviceAccountName: {{ template "agent.serviceAccountName" .}}
-{{- if .Values.priorityClassName }}
-      priorityClassName: "{{ .Values.priorityClassName }}"
+{{- if or .Values.priorityClassName .Values.gke.autopilot }}
```
mavimo: use `{{- include "agent.gke.autopilot" . }}` to check for autopilot using the helper.

charts/agent/templates/gkeautopilotpriorityclass.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if and .Values.gke.autopilot (not .Values.slim.enabled) }}
```
mavimo: use `{{- include "agent.gke.autopilot" . }}` to check for autopilot using the helper.

post-change-3.yaml
```diff
@@ -0,0 +1,469 @@
+---
```
aroberts87: Was this file intended to be committed?

pre-change-3.yaml
```diff
@@ -0,0 +1,486 @@
+---
```
aroberts87: Same as above, was this file intended to be committed?

charts/agent/templates/gkeautopilotpriorityclass.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if and (include "agent.gke.autopilot" .) (not .Values.slim.enabled) .Values.gke.createPriorityClass }}
```
aroberts87: Should we remove the constraint around slim not being enabled now?

charts/agent/templates/deployment.yaml
```diff
@@ -105,28 +102,13 @@ spec:
...
-              readOnly: true
-            {{- else }}
-            - mountPath: /host/etc
-              name: etc-vol
-              readOnly: true
-            {{- end }}
+          {{- /* Always requested */}}
             - mountPath: /host/dev
               name: dev-vol
+              readOnly: false
```
aroberts87: What is the reason for this change?

charts/agent/templates/daemonset.yaml
```diff
@@ -72,32 +72,29 @@ spec:
...
       {{- end }}
       {{- if .Values.slim.enabled }}
       initContainers:
         - name: sysdig-agent-kmodule
           image: {{ template "agent.image.kmodule" . }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           securityContext:
-            capabilities:
-              drop:
-                - ALL
```
aroberts87, line 85: What is the purpose of this change?

charts/agent/templates/daemonset.yaml
```diff
@@ -112,45 +109,51 @@ spec:
               value: {{ .Values.proxy.noProxy | default .Values.global.proxy.noProxy }}
           {{- end }}
           volumeMounts:
+          {{- /* Slim = true, Autopilot = false */}}
+          {{- if not (include "agent.gke.autopilot" .) }}
```
aroberts87, line 113: We're not checking the value of `slim.enabled` here, so won't this be picked in both cases?

charts/agent/templates/daemonset.yaml
```diff
@@ -112,45 +109,51 @@ spec:
...
         - name: sysdig
           image: {{ template "agent.image" . }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           resources:
-{{ include "agent.resources" . | indent 12 }}
+            {{- include "agent.resources" . | nindent 12 }}
           securityContext:
-            capabilities:
-              drop:
-                - ALL
```
aroberts87, line 153: What is the reason for this change?

charts/agent/templates/daemonset.yaml
mavimo: We're specifying this need to run as a privileged container, so the capabilities drop won't have any effect.

charts/agent/templates/deployment.yaml
mavimo: Is the default value, I added it supposing that make it more clear, but let's remove it 😅

charts/agent/templates/daemonset.yaml
mavimo: Is by default since the init container is available only in slim mode (see [line 77](https://github.com/sysdiglabs/charts/pull/975/files#diff-18888d7372d81c06ea1ec38c2606ae16bf04d0248bdb8ad475cbc5cc22bbc364R77))

charts/agent/templates/daemonset.yaml
mavimo: We're specifying this need to run as a privileged container, so the capabilities drop won't have any effect.

PR #985, opened 2023-03-13
chore(registry-scanner): filters review

wideawakening:
- maxWorkers, default from 1 to 5 + test
- maxTagsPerRepository, default to 5 + test
- maxAgeDays, default to 90 + test
- maxRepositoriesPerRegistry, default 500 + test


charts/registry-scanner/templates/configmap.yaml
```diff
@@ -80,8 +80,9 @@ data:
       exclude:
         {{- toYaml .Values.config.filter.exclude | nindent 8 }}
       {{- end }}
-      maxAgeDays: {{ default 0 .Values.config.filter.maxAgeDays }}
-      maxTagsPerRepository: {{ default 0 .Values.config.filter.maxTagsPerRepository }}
+      maxAgeDays: {{ .Values.config.filter.maxAgeDays }}
+      maxTagsPerRepository: {{ .Values.config.filter.maxTagsPerRepository }}
+      maxRepositories: {{ .Values.config.filter.repositoriesPerRegistry }}
```
tembleking: Please let's try to make consistent the key and the value, otherwise we are going to be confusing both.

tembleking (changes requested):

charts/registry-scanner/templates/configmap.yaml
tembleking:
```suggestion
      maxRepositories: {{ .Values.config.filter.maxRepositories }}
```

wideawakening: resolved by changing both key and value to `maxRepositoriesPerRegistry`

tembleking (approved):

PR #994, opened 2023-03-16
feat(registry-scanner): add extraEnvVars

wideawakening:


charts/registry-scanner/tests/job_test.yaml
```diff
@@ -19,3 +19,19 @@ tests:
       - equal:
           path: metadata.name
           value: foo
+#  - it: sets extraEnvVars on job
```
wideawakening, line 22: non-working test :[
any ideas?

charts/registry-scanner/tests/job_test.yaml
```diff
@@ -19,3 +19,19 @@ tests:
...
+#      - isNotEmpty:
+#          path: spec.template.spec.containers[0]
+#      - equal:
+#          path: spec.template.spec.containers[0].env["FOO"]
+#          value: fff
+#      - contains:
+#          path: spec.template.spec.containers[0].env
+#          content:
+#            name: FOO
+#            value: bar
```
wideawakening: non-working test :[
any ideas?

aroberts87: Hey @wideawakening! I replied in Slack, but for visibility here it seems like that test wasn't working because the `scanOnStart.enabled` parameter was not being set to true, and thus the job.tpl template was never being rendered as it is reliant on that flag being set to [true](https://github.com/sysdiglabs/charts/blob/master/charts/registry-scanner/templates/job.yaml#L1), but it is defaulted to [false](https://github.com/sysdiglabs/charts/blob/master/charts/registry-scanner/values.yaml#L199). Once you add that into the `set` block I believe the test should run as expected.

wideawakening: fixed! thanks for the extra 👁️

hayk99 (approved):

PR #996, opened 2023-03-17
chore(admission-controller,agent,kspm-collector,node-analyzer,rapid-response,registry-scanner,sysdig-deploy): update helm unittest plugin

mavimo:

 - Update the helm unittest plugin
 - Some README.md refactoring (formatting)
 - Add test folder for each chart
 - Fix broken tests related to the upgrade

## Checklist


tembleking (approved):
Thank you for the update

Makefile
```diff
@@ -1,20 +1,23 @@
...
-deps-helm:
-	helm plugin install https://github.com/quintush/helm-unittest  || true
-
-test-unit-all: deps-helm test-registry-scanner
+deps-unittest:
+	@helm plugin install https://github.com/helm-unittest/helm-unittest --version=0.3.0 || true
 
-test-registry-scanner:
-	helm unittest --helm3 ./charts/registry-scanner
+unittest: deps-unittest
```
wideawakening, line 19: non blocking feedback.

i prefer the makefile to be the main orchestrator for both pre-commit and github worklows to invoke it.
also, any reason on why the makefile has a generic  helm unittest, but the github workflow one invocation per chart?

wideawakening (approved):

Makefile
mavimo: @wideawakening I agree on having a single source of through and not having different definitions (e.g.: Makefile Vs GH Actions Vs ...) but not sure what will be the unified SSoT we are going to use here (I think should be a discussion that we need to start soon :D ). I'll keep you in the loop if you're interested in the discussion 🙌

PR #998, opened 2023-03-21
feat(agent): Expand /var/run volume mount to include entire /var

lrishi:
Agent requires access to container fs for internal state validation and feature support. On some platforms, these directories might reside outside of /var/run in /var/<vendor-specific> directories.

## Checklist


lrishi:
Tracked via https://github.com/sysdiglabs/charts/pull/1009

PR #1009, opened 2023-03-26
feat(agent): Expand /var/run volume mount to include entire /var

lrishi:
Agent requires access to container fs for internal state validation and feature support. On some platforms, these directories might reside outside of /var/run in /var/<vendor-specific> directories.

## Checklist

### DIFF
--- a/charts/agent/templates/deployment.yaml
+++ b/charts/agent/templates/deployment.yaml
@@ -122,8 +122,10 @@ spec:
               readOnly: true
             - mountPath: /host/run
               name: run-vol
-            - mountPath: /host/var
-              name: var-vol
+            - mountPath: /host/var/run
+              name: varrun-vol
+            - mountPath: /host/var/lib
+              name: varlib-vol
             - mountPath: /dev/shm
               name: dshm
             - mountPath: /opt/draios/etc/kubernetes/config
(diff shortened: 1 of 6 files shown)


charts/agent/templates/deployment.yaml
```diff
@@ -184,9 +184,9 @@ spec:
             type: ""
           name: run-vol
         - hostPath:
-            path: /var/run
+            path: /var
```
mavimo: The deployment is intended only for collecting k8s infos, do we need to get access to the whole `/var` folder also in this case?

lrishi: If that is the case, then no, but since its agent deployment, my understanding is that both are 1/1 mapped, except for Deployment/Daemonset, but I might be wrong. There is some value to keeping both in sync, but I'm happy to remove this part, if you suggest so.

mavimo: @lrishi IMHO exposing information that are not relevant into container may be not needed, at the same time it may increase maintenance complexity; let ask for a 3rd opinion from @aroberts87

aroberts87: Hey @lrishi - I agree with @mavimo and I believe you can remove the change to map all of `/var` into the Deployment agents. Those agents are running in `nodriver` mode and do not send syscall metrics.

mavimo:
The scope of this PR is changed, during tests some issue emerged.

Mounting the whole `/var` folder may create issues related to symlinks (e.g.: `/var/run -> /run` we ending up by exposing in the `/host/var/run` the `/run` folder in the container instead of the `/run` folder of the host).
Since @lrishi report that only the `/var/lib` folder is required to be exposed in the `/host/var/lib` folder we limit the mount of this specific folder in order to limit issues.

mavimo (approved):

PR #1039, opened 2023-04-07
fix(agent,sysdig-deploy): Reduce Agent mode misconfiguration from fail to NOTES output

aroberts87:
The introduction of the monitor.enabled flag introduced a breaking change for users explicitly putting the Agent into secure or secure_light mode by informing the user that putting the Agent into one of those two modes while running the Sysdig Monitor components is not advised. This change reduces the intentional rendering fail into a suggestion in the NOTES.txt output.

## Checklist

### DIFF
--- a/charts/agent/tests/monitor_enable_test.yaml
+++ b/charts/agent/tests/monitor_enable_test.yaml
@@ -1,6 +1,7 @@
 suite: Test enabling/disabling monitor features
 templates:
   - templates/configmap.yaml
+  - templates/NOTES.txt
 tests:
   - it: Set chart defaults (monitor.enabled=true)
     asserts:
@@ -11,6 +12,7 @@ tests:
           pattern: |-
             app_checks_enabled: false
         not: true
+    template: templates/configmap.yaml
 
   - it: Set monitor.enabled=true
     set:
@@ -24,6 +26,7 @@ tests:
           pattern: |-
             app_checks_enabled: false
         not: true
+    template: templates/configmap.yaml
 
   - it: Set monitor.enabled=false
     set:
@@ -42,6 +45,7 @@ tests:
               enabled: false
             statsd:
               enabled: false
+    template: templates/configmap.yaml
 
   - it: Test monitor.enabled when specifying secure_light mode
     set:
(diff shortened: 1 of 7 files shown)


charts/sysdig-deploy/tests/notes_test.yaml
```diff
@@ -72,3 +72,37 @@ tests:
     asserts:
       - failedTemplate:
           errorMessage: "raw: global.sysdig.region=ap3 provided is not recognized."
+
+  - it: Test agent.monitor.enabled when specifying Agent secure_light mode
+    set:
+      agent:
+        monitor:
+          enabled: true
```
mavimo, line 80: If we are testing object creation, we must also test the other way round (test that no objects are created if `false'). 🙃

charts/sysdig-deploy/tests/notes_test.yaml
```diff
@@ -72,3 +72,37 @@ tests:
...
+            The agent.monitor.enabled parameter is true while using agent.sysdig.settings.feature.mode to put the Agent into either
+            secure or secure_light mode. Please set agent.monitor.enabled=false to ensure all Sysdig Monitor components are disabled
+            when running the Agent in secure or secure_light modes.
+    template: templates/NOTES.txt
+
+  - it: Test agent.monitor.enabled when specifying Agent secure mode
+    set:
+      agent:
+        monitor:
+          enabled: true
```
mavimo, line 114: Same as before

charts/agent/tests/monitor_enable_test.yaml
```diff
@@ -52,8 +56,12 @@ tests:
           feature:
             mode: secure_light
     asserts:
-      - failedTemplate:
-          errorMessage: 'Cannot set monitor.enabled=true when sysdig.settings.feature.mode is `secure` or `secure_light`'
+      - matchRegexRaw:
```
mavimo, line 59: Test also the `notMatchRegexRaw` when the monitor.enabled = false. 🙏

charts/agent/tests/monitor_enable_test.yaml
```diff
@@ -64,5 +72,9 @@ tests:
           feature:
             mode: secure
     asserts:
-      - failedTemplate:
-          errorMessage: 'Cannot set monitor.enabled=true when sysdig.settings.feature.mode is `secure` or `secure_light`'
+      - matchRegexRaw:
```
mavimo, line 91: same as before

mavimo (changes requested):

mavimo (approved):

PR #1040, opened 2023-04-07
fix(admission-controller,agent,node-analyzer,rapid-response,registry-scanner,sysdig): Improve KubeVersion Comparisons

aroberts87:
In various places across many charts we do compatibility checks against exact Kubernetes versions. To do so, the charts had been casting the raw value of `.Capabilities.KubeVersion.Minor` to an int like the following:

```
(lt (int .Capabilities.KubeVersion.Minor) 25)
```

The bug here is that certain Kubernetes flavors (EKS, microk8s, etc.) will sometimes put a `+` at the end of the minor version. That `+` was causing the cast to an int to silently fail and return a zero, which was causing failure by attempting to deploy constructs that are not supported on incorrect cluster versions.

This change guards against the `+` suffix as well as abstracts away the version check logic into another helper function that can be more easily adjusted if necessary.

## Checklist

### DIFF
--- a/charts/admission-controller/tests/psp_test.yaml
+++ b/charts/admission-controller/tests/psp_test.yaml
@@ -1,0 +1,35 @@
+suite: PSP create test
+templates:
+  - templates/scanner/psp.yaml
+  - templates/scanner/role.yaml
+  - templates/scanner/rolebinding.yaml
+tests:
+  - it: Ensure PSP is created on k8s <1.25
+    capabilities:
+      majorVersion: 1
+      minorVersion: 24
+    set:
+      scanner:
+        psp:
+          create: true
+    asserts:
+      - containsDocument:
+          apiVersion: policy/v1beta1
+          kind: PodSecurityPolicy
+        template: templates/scanner/psp.yaml
+      - containsDocument:
+          apiVersion: rbac.authorization.k8s.io/v1
+          kind: Role
+        template: templates/scanner/role.yaml
+      - containsDocument:
+          apiVersion: rbac.authorization.k8s.io/v1
+          kind: RoleBinding
+        template: templates/scanner/rolebinding.yaml
+
+  - it: Ensure PSP is not created on k8s >=1.25
+    capabilities:
+      majorVersion: 1
+      minorVersion: 25
+    asserts:
+      - hasDocuments:
+          count: 0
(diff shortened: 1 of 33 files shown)


charts/admission-controller/tests/psp_test.yaml
```diff
@@ -0,0 +1,35 @@
...
+          kind: RoleBinding
+        template: templates/scanner/rolebinding.yaml
+
+  - it: Ensure PSP is not created on k8s >=1.25
+    capabilities:
+      majorVersion: 1
+      minorVersion: 25
+    asserts:
+      - hasDocuments:
+          count: 0
```
mavimo, line 35: Can we add also a check for EKS?
Something like (to be validated)
```yaml
  - it: Ensure PSP is not created on EKS k8s >=1.25
    capabilities:
      majorVersion: 1
      minorVersion: "25+"
    asserts:
      - hasDocuments:
          count: 0
  - it: Ensure PSP is created on EKS k8s <1.25
    capabilities:
      majorVersion: 1
      minorVersion: "24+"
    asserts:
      - hasDocuments:
          count: 1
```

charts/sysdig/templates/_helpers.tpl
```diff
@@ -295,3 +295,16 @@ true
 true
 {{- end -}}
 {{- end -}}
+
+{{/* Returns string 'true' if the cluster's kubeVersion is less than the parameter provided, or nothing otherwise
+     Use like: {{ include "sysdig.kubeVersionLessThan" (dict "root" . "major" "<kube_major_to_compare>" "minor" "<kube_minor_to_compare>") }}
```
mavimo:
```suggestion
     Use like: {{ include "sysdig.kubeVersionLessThan" (dict "root" . "major"<kube_major_to_compare> "minor" <kube_minor_to_compare>) }}
```
major and minor should be number, adding quotes might sound like we need to indicate them as string but then we are not casting them as integer into template..

charts/admission-controller/tests/psp_test.yaml
aroberts87: Done! 🙂

charts/sysdig/templates/_helpers.tpl
aroberts87: All set. 🙂

mavimo (approved):

PR #1041, opened 2023-04-10
feat(agent): Move Agent lease permissions

aroberts87:
The lease permissions had been defined in the ClusterRole, but that was too permissive. This change creates a new Role and RoleBinding for the lease permissions so that they are scoped to the Agent's namespace.

## Checklist

### DIFF
--- a/charts/agent/tests/role_test.yaml
+++ b/charts/agent/tests/role_test.yaml
@@ -37,3 +37,15 @@ tests:
             name: RELEASE-NAME-agent
             namespace: NAMESPACE
     template: templates/rolebinding.yaml
+
+  - it: Role and RoleBinding are not created when rbac.create is set to false
+    set:
+      rbac:
+        create: false
+    asserts:
+      - hasDocuments:
+          count: 0
+        template: templates/role.yaml
+      - hasDocuments:
+          count: 0
+        template: templates/rolebinding.yaml
--- a/charts/agent/tests/role_test.yaml
+++ b/charts/agent/tests/role_test.yaml
@@ -1,0 +1,39 @@
+suite: Agent Role Tests
+templates:
+  - templates/role.yaml
+  - templates/rolebinding.yaml
+tests:
+  - it: Role and RoleBinding created when rbac.create is set
+    set:
+      rbac:
+        create: true
+    asserts:
+      - containsDocument:
+          name: RELEASE-NAME-agent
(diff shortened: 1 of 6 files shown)


charts/agent/tests/role_test.yaml
```diff
@@ -0,0 +1,39 @@
+suite: Agent Role Tests
+templates:
+  - templates/role.yaml
+  - templates/rolebinding.yaml
+tests:
+  - it: Role and RoleBinding created when rbac.create is set
+    set:
+      rbac:
+        create: true
```
mavimo, line 9: If we are testing object creation, we must also test the other way round (test that no objects are created if `false') 😺

charts/agent/tests/role_test.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - containsDocument:
+          name:  RELEASE-NAME-agent
+          apiVersion: rbac.authorization.k8s.io/v1
+          kind: RoleBinding
+        template: templates/rolebinding.yaml
+
+  - it: RoleBinding linked with Role and ServiceAccount
+    set:
+      rbac:
+        create: true
```
mavimo, line 25: Same as before

mavimo (changes requested):

charts/agent/tests/role_test.yaml
aroberts87: Hey @mavimo - The test you suggested above will ensure that the Role and RoleBinding are not created. This test is validating that when the RoleBinding is created the content is what is expected. If we have already validated that the  RoleBinding is not created, I'm not sure what the negative test is in this scenario. Am I missing something?

mavimo: Right, sorry, we need to test "don't create role & role binding" when `rbac.create = false`

aroberts87: Are we okay to merge this PR then?

mavimo: I add the test for `rbac.create = false` in [d30ccee39811215b1bfc0cd6739fdeaa9e6859f6](https://github.com/sysdiglabs/charts/pull/1041/files#diff-f292da8564fc3a39cb5aaf5a4f2d61861163037583f1646aaaf3d47d49318016R41-R51)

mavimo (approved):

PR #1052, opened 2023-04-19
fix(sysdig-deploy): nil pointer evaluating interface {}.autopilot

dark-vex:

When trying to deploy for example just the admission controller via `sysdig-deploy` with `agent.enabled` set to false, it fails with the following error message:
```
helm upgrade --install sysdig -f eks.yaml sysdig/sysdig-deploy
Error: UPGRADE FAILED: template: sysdig-deploy/templates/NOTES.txt:13:18: executing "sysdig-deploy/templates/NOTES.txt" at <.Values.agent.gke.autopilot>: nil pointer evaluating interface {}.autopilot
```

Sample config:
```
global:
  clusterConfig:
    name: "eks-test"
  sysdig:
    accessKey: "KEY"
    region: "us2"
    secureAPIToken: "TOKEN"
  kspm:
    deploy: false
  
agent:
  enabled: false

nodeAnalyzer:
  enabled: false

admissionController:
  enabled: true
  scanner:
    enabled: false
```

## Checklist

### DIFF
--- a/charts/sysdig-deploy/templates/NOTES.txt
+++ b/charts/sysdig-deploy/templates/NOTES.txt
@@ -11,27 +11,27 @@ Usage of global tags will override any local tags you might be using.
 {{- end }}
 
 {{- if .Values.agent.enabled }}
-{{- if (or .Values.agent.gke.autopilot .Values.global.gke.autopilot) }}
-    {{- if hasKey .Values.agent.sysdig.settings "drift_killer" }}
-        {{- if hasKey .Values.agent.sysdig.settings.drift_killer "enabled" }}
-            {{- if .Values.agent.sysdig.settings.drift_killer.enabled }}
+    {{- if (or .Values.agent.gke.autopilot .Values.global.gke.autopilot) }}
+        {{- if hasKey .Values.agent.sysdig.settings "drift_killer" }}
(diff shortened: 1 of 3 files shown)


charts/sysdig-deploy/templates/NOTES.txt
```diff
@@ -10,6 +10,7 @@ Links for your deployment:
 Usage of global tags will override any local tags you might be using.
 {{- end }}
 
+{{- if .Values.agent.enabled }}
```
mavimo, line 13: @dark-vex thx for the fix, WDYT to add this check in the same if condition in the line below, like:
```suggestion
{{- if (and .Values.agent.enabled (or .Values.agent.gke.autopilot .Values.global.gke.autopilot)) }}
```

dark-vex: Hey @mavimo I did that before opening the PR but then it fails on this other two lines:
https://github.com/sysdiglabs/charts/blob/master/charts/sysdig-deploy/templates/NOTES.txt#L25
https://github.com/sysdiglabs/charts/blob/master/charts/sysdig-deploy/templates/NOTES.txt#L31
This is why I did `{{- if .Values.agent.enabled }}` for the entire block instead of adding `and .Values.agent.enabled` for each `if` entry

mavimo (approved):

PR #1056, opened 2023-04-19
feat(cluster-scanner): init chart with minimal templates and configuration

ProvoK:
## Checklist


charts/cluster-scanner/templates/_helpers.tpl
```diff
@@ -0,0 +1,125 @@
...
+{{- end }}
+{{- end }}
+
+{{/*
+Generates configmap data for mode-specific values
+*/}}
+{{- define "cluster-scanner.modeConfig" -}}
+{{- if eq .Values.global.scannerMode "local" }}
+rsi_mode: "sitting"
+{{- else }}
```
ProvoK: if ... else, otherwhise return error?

charts/cluster-scanner/templates/_helpers.tpl
```diff
@@ -0,0 +1,125 @@
...
+{{- default "default" .Values.serviceAccount.name }}
+{{- end }}
+{{- end }}
+
+{{/*
+Generates configmap data for mode-specific values
+*/}}
+{{- define "cluster-scanner.modeConfig" -}}
+{{- if eq .Values.global.scannerMode "local" }}
```
FedericoFeresini: Why don't we stick to `sitting` and `mcm` also for the scannerMode?

charts/cluster-scanner/templates/deployment.yaml
```diff
@@ -0,0 +1,398 @@
...
+          image: "{{ .image.repository }}:{{ .image.tag }}"
+          imagePullPolicy: {{ .image.pullPolicy }}
+          ports:
+            - name: metrics
+              containerPort: {{ .ports.metrics }}
+              protocol: TCP
+            - name: probes
+              containerPort: {{ .ports.probes }}
+              protocol: TCP
+          env:
```
mavimo, line 48: Is it possible to get rid from ENVs and read informations from  a file? This is specially useful in case of information that should not be disclosed (access keys, password, ..) like in `SYSDIG_ACCESS_KEY`, `ANALYZER_CACHE_REDIS_PASSWORD`, ...

charts/cluster-scanner/templates/rbac.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - "statefulsets"
+    - "pods"
+    - "cronjobs"
+    - "jobs"
+    - "nodes"
+    - "secrets"
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["", "coordination.k8s.io"]
+  resources:
+    - "leases"
```
mavimo: Since the lease is going to be used in the bootstrap and is a namespaced resource we should avoid to create it as ClusterRole level and limit to the namespace we are going to use with a Role + RoleBinding (see [#1041](https://github.com/sysdiglabs/charts/pull/1041))

charts/cluster-scanner/values.yaml
```diff
@@ -0,0 +1,124 @@
...
+    probes: 7001
+  resources:
+    limits:
+      cpu: "1"
+      memory: 350Mi
+    requests:
+      cpu: "150m"
+      memory: 350Mi
+
+imagePullSecrets: []
```
mavimo: I suppose the image pull secret should be at component level (ise / rsi / ...)

charts/cluster-scanner/values.yaml
```diff
@@ -0,0 +1,124 @@
...
+
+replicaCount: 2
+
+# Params to manage leader election of rsi component.
+# Leader election is implemented leveraging the native capabilities of Kubernetes
+# see: https://kubernetes.io/blog/2016/01/simple-leader-election-with-kubernetes/
+# leaderElectionLeaseNameOverride: ""
+# leaderElectionLeaseNamespaceOverride: ""
+
+rsi:
```
mavimo: Can we consider to use the full name here? I think it become more understandable for a user then using the acronym.. (same for `ise` a few line below)

charts/cluster-scanner/templates/_helpers.tpl
```diff
@@ -0,0 +1,127 @@
...
+rsi_mode: "sitting"
+{{- else if eq .Values.global.scannerMode "multi" }}
+rsi_mode: "mcm"
+{{- else }}
+{{- fail "invalid scannerMode, accepted options [local, multi]" }}
```
mavimo: This can be enforced at the chart level creating the proper `values.schema.json` without adding checks in the template (simplify templates and making them more readable)

charts/cluster-scanner/values.yaml
```diff
@@ -0,0 +1,124 @@
...
+tolerations: []
+
+affinity: {}
+
+js:
+  user: "default-user"
+  # Default uses a random password for each installation.
+  # password: ""
+
+cache:
```
mavimo: This is `ise` specific config, WDYT to move it under the `ise` key?

charts/cluster-scanner/values.yaml
```diff
@@ -0,0 +1,124 @@
...
+selectorLabels: {}
+
+
+nodeSelector: {}
+
+tolerations: []
+
+affinity: {}
+
+js:
```
mavimo:
```suggestion
natsJs:
```

charts/cluster-scanner/values.yaml
```diff
@@ -0,0 +1,124 @@
...
+  sysdigSkipVerifyCertificate: false
+
+  scannerMode: "local" # or "multi"
+  loggingLevel: "INFO"
+
+eveEnabled: false
+eveIntegrationEnabled: false
+rootNamespace: "kube-system"
+
+multicluster:
```
mavimo: We are using camelCAse for all the other config, I suggest to keep consistency.
```suggestion
multiCluster:
```

charts/cluster-scanner/values.yaml
```diff
@@ -0,0 +1,124 @@
...
+  # kubeconfigSecretName: ""
+
+localcluster:
```
mavimo: We are using camelCase for all the other config, I suggest to keep consistency.

```suggestion
localCluster:
```

charts/cluster-scanner/templates/deployment.yaml
```diff
@@ -0,0 +1,398 @@
...
+      {{- with .Values.rsi }}
+          image: "{{ .image.repository }}:{{ .image.tag }}"
+          imagePullPolicy: {{ .image.pullPolicy }}
+          ports:
+            - name: metrics
+              containerPort: {{ .ports.metrics }}
+              protocol: TCP
+            - name: probes
+              containerPort: {{ .ports.probes }}
+              protocol: TCP
```
mavimo: Unless they need to be accessed from other tools, probes port don't need to be exposed

charts/cluster-scanner/templates/deployment.yaml
```diff
@@ -0,0 +1,398 @@
...
+      {{- with .Values.ise }}
+          image: "{{ .image.repository }}:{{ .image.tag }}"
+          imagePullPolicy: {{ .image.pullPolicy }}
+          ports:
+            - name: metrics
+              containerPort: {{ .ports.metrics }}
+              protocol: TCP
+            - name: probes
+              containerPort: {{ .ports.probes }}
+              protocol: TCP
```
mavimo: Unless they need to be accessed from other tools, probes port don't need to be exposed

charts/cluster-scanner/values.yaml
```diff
@@ -0,0 +1,124 @@
...
+# Declare variables to be passed into your templates.
+
+global:
+  sysdigHost: ""
+  sysdigAccessKey: ""
+  clusterName: ""
+  sysdigSkipVerifyCertificate: false
+
+  scannerMode: "local" # or "multi"
+  loggingLevel: "INFO"
```
mavimo: Can we try to keep the same structure we have in the main chart here?

https://github.com/sysdiglabs/charts/blob/master/charts/sysdig-deploy/values.yaml#L1-L17

charts/cluster-scanner/values.yaml
```diff
@@ -0,0 +1,124 @@
...
+
+eveEnabled: false
+eveIntegrationEnabled: false
```
mavimo: This should be grouped in a single component as `rsi` and `ise`?

Can we -at least- try to have a nested structure for `eve` like:
```
eve:
    enabled: false
    integration:
        enabled: false
```

charts/cluster-scanner/values.yaml
mavimo: This seels to be specific for rsi, can we move it as config in that section?

charts/cluster-scanner/values.yaml
mavimo: Seems not used, can we remove it?

charts/cluster-scanner/README.md
```diff
@@ -0,0 +1,5 @@
+# Cluster Scanner
+
+:warning: This chart is currently under construction :warning:
+
+Please don't use it unless you're really sure about what you are doing.
```
mavimo: please migrate to the auto generated readme, see [admission-controller](https://github.com/sysdiglabs/charts/blob/master/charts/admission-controller/README.tpl) as example.

mavimo (changes requested):

charts/cluster-scanner/templates/_helpers.tpl
ProvoK: easier to understand from non-dev personas

charts/cluster-scanner/templates/deployment.yaml
ProvoK: atm no, there is no such capability in the services. We could try to add this effort later on 

cc @michele-mangili @guidobonomi

charts/cluster-scanner/values.yaml
ProvoK: I was thinking of keeping cache config global across containers in case `rsi` needs it, and also to avoid exposing that level of detail to end user (which just want to enable the distributed cache)

makes sense to you? @mavimo

charts/cluster-scanner/values.yaml
ProvoK: it's actually a placeholder: the current service implementation does not support it, but it will

cc @michele-mangili

charts/cluster-scanner/values.yaml
ProvoK: I was following Helm Chart [suggested best practices](https://helm.sh/docs/chart_best_practices/values/#flat-or-nested-values) about values and avoiding unnecessary nesting

if you feel like it needs nesting, I can surely change it

---

regarding grouping/putting config inside `rsi` and `ise`, I was trying to put in there the bare minimum set of parameters that needs to be scoped to a single container:
end-users won't care about which container does what, but just enable/disable EVE, or the cache, etc.

So, unless strictly required (eg: image name/tag) I would avoid requiring end-user to know how we split responsibilities into the applications/containers.

Does it make sense to you? @mavimo

charts/cluster-scanner/values.yaml
ProvoK: pullsecrets are defined once per pod, it's not at the container level

charts/cluster-scanner/values.yaml
ProvoK: @mavimo I'll do, but when using `custom` reagion, which parameter is supposed to specify the host/baseURL for the Sysdig backend?

charts/cluster-scanner/values.yaml
ProvoK: same reasons as per https://github.com/sysdiglabs/charts/pull/1056#discussion_r1173869100

charts/cluster-scanner/values.yaml
ProvoK: I'll add a `global.sysdig.apiHost`, possibly we can change that

charts/cluster-scanner/Makefile
```diff
@@ -0,0 +1,3 @@
+.PHONY: test
+test:
+	helm unittest -3 .
```
mavimo:
```suggestion
	helm unittest .
```

the flag `-3` is not supported with the unittest plugin (>=0.3.x)

charts/cluster-scanner/.helmignore
```diff
@@ -0,0 +1,23 @@
...
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
```
mavimo, line 23: Add also the `ci` and `tests` folders

charts/cluster-scanner/values.yaml
```diff
@@ -0,0 +1,136 @@
...
+  ports:
+    metrics: 25000
+    probes: 7000
+  resources:
+    limits:
+      cpu: "1"
+      memory: 350Mi
+    requests:
+      cpu: "350m"
+      memory: 350Mi
```
mavimo: Is it correct to set request & limit in the chart as default? How do we get this values?

charts/cluster-scanner/values.yaml
```diff
@@ -0,0 +1,136 @@
...
+  ports:
+    metrics: 25001
+    probes: 7001
+  resources:
+    limits:
+      cpu: "1"
+      memory: 350Mi
+    requests:
+      cpu: "150m"
+      memory: 350Mi
```
mavimo: Is it correct to set request & limit in the chart as default? How do we get this values?

charts/node-analyzer/templates/_helpers.tpl
```diff
@@ -230,14 +230,14 @@ nodeAnalyzer agentConfigmapName
...
-{{- else if or .Values.secure.vulnerabilityManagement.newEngineOnly (and (hasKey .Values.nodeAnalyzer.hostScanner "deploy") .Values.nodeAnalyzer.hostScanner.deploy) -}}
+{{- if and (hasKey ((.Values.nodeAnalyzer).hostScanner) "deploy") (not .Values.nodeAnalyzer.hostScanner.deploy ) }}
+{{- else if or ((.Values.secure).vulnerabilityManagement).newEngineOnly (and (hasKey ((.Values.nodeAnalyzer).hostScanner) "deploy") .Values.nodeAnalyzer.hostScanner.deploy) -}}
 true
 {{- end -}}
 {{- end -}}
 
 {{- define "nodeAnalyzer.deployRuntimeScanner" -}}
-{{- if or .Values.secure.vulnerabilityManagement.newEngineOnly (not (hasKey .Values.nodeAnalyzer.runtimeScanner "deploy")) .Values.nodeAnalyzer.runtimeScanner.deploy }}
+{{- if or ((.Values.secure).vulnerabilityManagement).newEngineOnly (not (hasKey ((.Values.nodeAnalyzer).runtimeScanner) "deploy")) .Values.nodeAnalyzer.runtimeScanner.deploy }}
```
mavimo: Why do we need parenthesis here? `(.Values.secure).vulnerabilityManagement` 🤔

charts/cluster-scanner/values.yaml
ProvoK: @mavimo For this specific case, it is.

These components will require a minimum set of resources for working (for pulling image layers, using the vulnerability database & so on)
The amount has been calculated in the last months of continuous improvements and monitoring (and is subject to change in the future).

If you look at that node-analyzer.runtimeScanner is doing, it's the same

charts/cluster-scanner/values.yaml
ProvoK: same as per https://github.com/sysdiglabs/charts/pull/1056#discussion_r1187390062

charts/node-analyzer/templates/_helpers.tpl
ProvoK: it's a safety check: if `.Values` does not contain `secure` property, it won't explode and default to base value (empty string I believe).

michele-mangili (changes requested):
Marking as Request changes to prevent unintended merge to `master` (need confirmation from Product to release the `cluster-scanner` component).

charts/cluster-scanner/templates/deployment.yaml
michele-mangili: We have a task in the product backlog to address this in upcoming releases https://sysdig.atlassian.net/browse/SSPROD-24679 . It may also be beneficial to reduce the number of env vars we need to set, and simplify the overall configuration of the cluster-scanner.

charts/cluster-scanner/values.yaml
michele-mangili: Marking this as resolved, since the `cache` configuration is now specific to the `ise` itself. See https://github.com/sysdiglabs/charts/blob/80bb5bbbde270249bc4bb7bca5413e49e117bc1b/charts/cluster-scanner/values.yaml#L163

charts/cluster-scanner/values.yaml
michele-mangili: Marking this as resolved since we moved this configuration under the `runtimeStatusIntegrator` itself. See: https://github.com/sysdiglabs/charts/blob/80bb5bbbde270249bc4bb7bca5413e49e117bc1b/charts/cluster-scanner/values.yaml#L81

charts/cluster-scanner/values.yaml
michele-mangili: We have finalized the implementation and the `localCluster` config is supported now. See https://github.com/sysdiglabs/charts/blob/80bb5bbbde270249bc4bb7bca5413e49e117bc1b/charts/cluster-scanner/values.yaml#L84

charts/cluster-scanner/values.yaml
michele-mangili: We have updated it see: https://github.com/sysdiglabs/charts/blob/f3c19117aae903614c4d7aa59a8848db1ca81f06/charts/cluster-scanner/values.yaml#L5-L26

michele-mangili (approved):

mavimo (approved):

PR #1060, opened 2023-04-24
fix(cluster-scanner): review fixes

ProvoK:
as per title


charts/cluster-scanner/templates/_helpers.tpl
```diff
@@ -71,10 +71,9 @@ Generates configmap data for mode-specific values
 {{- define "cluster-scanner.modeConfig" -}}
 {{- if eq .Values.global.scannerMode "local" }}
 rsi_mode: "sitting"
-{{- else if eq .Values.global.scannerMode "multi" }}
-rsi_mode: "mcm"
+{{/* NOTE: values.schema.json will rpevent any other non valid strings, so it's safe to assume it's "multi" */}}
```
mavimo:
```suggestion
{{/* NOTE: values.schema.json will prevent any other non valid strings, so it's safe to assume it's "multi" */}}
```

PR #1066, opened 2023-05-02
feat(cluster-scanner): bumped the RSI and ISE version and improved the chart

michele-mangili:
## Checklist


charts/cluster-scanner/templates/_helpers.tpl
```diff
@@ -91,7 +91,10 @@ rsi_js_consumer_subject_prefix: "analysis.sboms"
 rsi_js_consumer_max_in_flight: "256"
 rsi_js_consumer_ack_wait: "120s"
 rsi_js_consumer_max_deliver: "1"
+rsi_js_consumer_deliver_policy_all: "true"
 rsi_js_producer_subject_prefix: "analysis.requests"
+rsi_js_server_metrics_enable: "true"
```
ProvoK, line 96: if we expose metrics by default we should annotate our deployment resource with the "standard" labels (spec.template.metadata.annotations) which are 

```
        prometheus.io/scrape: "true"
        prometheus.io/port: "25000"
```

michele-mangili: Ok I have added them! Thank you!

ProvoK (approved):
LGTM!

PR #1079, opened 2023-05-09
fix(agent): Allow cointerface to be enabled in daemonset agents in delegated agent deployment

aroberts87:
There exist some use cases where having cointerface enabled on both the Deployment and DaemonSet Agents would be beneficial. This change removes the hard requirement that cointerface is disabled in the DaemonSet. Cointerface will remain disabled by default in the delegated agent deployments, but should the user specify `cointerface_enabled: true` in the agent chart's `sysdig.settings` block, cointerface will remain enabled in the DaemonSet Agents.

## Checklist

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -5,7 +5,7 @@ description: Sysdig Monitor and Secure agent
 type: application
 
 # currently matching sysdig 1.14.32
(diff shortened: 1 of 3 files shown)


mavimo (approved):

irozzo-1A (approved):
LGTM

PR #1081, opened 2023-05-09
chore(registry-scanner): orchestrator job label

wideawakening:
adds labels to orchestrator job `app.kubernetes.io/component:registry-scanner-orchestrator`
inner work jobs labels have been added pragmatically too in bumped registry-scanner version  `app.kubernetes.io/component:registry-scanner-worker`

### DIFF
--- a/charts/registry-scanner/templates/_helpers.tpl
+++ b/charts/registry-scanner/templates/_helpers.tpl
@@ -35,6 +35,7 @@ Common labels
 */}}
 {{- define "registry-scanner.labels" -}}
 helm.sh/chart: {{ include "registry-scanner.chart" . }}
+product-type: registry-scanner-orchestrator
 {{ include "registry-scanner.selectorLabels" . }}
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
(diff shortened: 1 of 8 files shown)


hayk99 (approved):

charts/registry-scanner/templates/_helpers.tpl
```diff
@@ -35,6 +35,7 @@ Common labels
 */}}
 {{- define "registry-scanner.labels" -}}
 helm.sh/chart: {{ include "registry-scanner.chart" . }}
+product-type: registry-scanner-orchestrator
```
mavimo: Why do we need this label? Should we consider to use one of the well-known label for that? https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

wideawakening: purpose is, we want to differentiate from orchestrator vs. worker job logs by label.

i like the conventions from k8s @mavimo , thx!
what about using
```
    app.kubernetes.io/name: registry-scanner
    app.kubernetes.io/component: orchestrator | worker
```

would that suit better?

wideawakening: addressed both on chart and inner component

PR #1083, opened 2023-05-10
feat: update mode names

ProvoK:
as per title


michele-mangili (approved):
LGTM!

I guess the `Error: INSTALLATION FAILED: timed out waiting for the condition` error is expected, given that the version we are using now does not support this new rsi mode.

ProvoK:
outdated

PR #1086, opened 2023-05-11
feat(cluster-scanner): version compatibility checks, added NOTES.txt and improved documentation

michele-mangili:
## Checklist


charts/cluster-scanner/templates/NOTES.txt
```diff
@@ -1 +1,9 @@
-TODO
+The Cluster Scanner component is being executed.
+
+It consists of the Runtime Status Integrator and Image Sbom Extractor components.
+
+After successful installation you can access the Cluster Scanner features as follows:
+
+Log in to Sysdig Secure and check that the features are working as expected.
```
ProvoK: @michele-mangili this will actually be rendered with values, as part of the chart (and could also get actual live data from the kubernetes cluster!!!)

see: https://helm.sh/docs/chart_template_guide/notes_files/

I'm saying this because we could generate the UI URL instead of telling how to reach it :)

ProvoK: plus, unrelated, please add that results won't be instantly available 🙏

michele-mangili: Sure, will do! I copy-pasted the notes we currently display for the node-analyzer.

FedericoFeresini: I don't know if this is a standard, if not I'd suggest that we do not print out a message asking the client to check out if everything is working as expected.

michele-mangili: @FedericoFeresini I copy-pasted the message we currently display for the node analyzer https://github.com/sysdiglabs/charts/blob/006d2fc36547a0ab8c2f1d6a26bea75ac93651a8/charts/node-analyzer/templates/NOTES.txt 

In that file we have exactly the same statement: `Log in to Sysdig Secure and check that the features are working as expected.`. So I guess our customers expect something like that.

ProvoK (approved):

PR #1093, opened 2023-05-12
feat(node-analyzer): support settings passed as numbers

### LINKED ISSUE #1092: Specifying runtimeScanner maxFileSizeAllowed as integer can result in failures
**Summary**

If users specify the `nodeAnalyzer.runtimeScanner.settings.maxFileSizeAllowed` setting as an integer, then the `quote` function sometimes converts it to a string correctly (for small numbers) or otherwise treats it as a floating-point number, which then causes the runtime scanner container to fail to start.

**Current behavior**

When `maxFileSizeAllowed` is given as an integer (rather than a string), an incorrect value appears in the resulting `ConfigMap`, which prevents the scanner from starting correctly.

**Desired behavior**

In both cases, the chart should be able to accept either a string or int64.

**Steps to reproduce**

1. Use the following values file and save it somewhere as `values.yaml`:
   ```yaml
   sysdig:
     accessKey: foo

   clusterName: foo

   nodeAnalyzer:
     deploy: true
     apiEndpoint: https://api.example.com
     runtimeScanner:
       deploy: true
       settings:
       maxFileSizeAllowed: 262144000
   ```

1. Run `helm template sysdig/sysdig --values=values.yaml` (it likely also works when running locally against this repo, but I have not checked)
1. Look at the output; for me, it looks like this:
   ```yaml
   # Source: sysdig/templates/runtimeScanner/runtime-scanner-configmap.yaml
   apiVersion: v1
   kind: ConfigMap
   metadata:
     name: release-name-sysdig-runtime-scanner
     labels:
       helm.sh/chart: sysdig-1.15.87
       app.kubernetes.io/instance: release-name
       app.kubernetes.io/version: "12.14.0"
       app.kubernetes.io/managed-by: Helm
       app: "sysdig-agent"
   data:
     api_endpoint: https://https://api.example.com
     cluster_name: foo
     debug: "false"
     eve_integration_enabled: "true"
     prom_port: "25001"
     analyzer.maxFileSizeAllowed: "2.62144e+08"
   ```

You can also try changing the `maxFileSizeAllowed` to a smaller number, and then `quote` behaves as we expect:

```yaml
# Source: sysdig/templates/runtimeScanner/runtime-scanner-configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: release-name-sysdig-runtime-scanner
  labels:
    helm.sh/chart: sysdig-1.15.87
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "12.14.0"
    app.kubernetes.io/managed-by: Helm
    app: "sysdig-agent"
data:
  api_endpoint: https://https://api.example.com
  cluster_name: foo
  debug: "false"
  eve_integration_enabled: "true"
  prom_port: "25001"
  analyzer.maxFileSizeAllowed: "2621"
```

I also verified that it is possible to reproduce this with the new `sysdig-deploy` chart using this values file:

```yaml
global:
  sysdig:
    accessKey: foo

agent:
  enabled: false

nodeAnalyzer:
  enabled: true
  clusterName: foo
  apiEndpoint: https://api.example.com
  secure:
    enabled: true
    vulnerabilityManagement:
      newEngineOnly: true
  nodeAnalyzer:
    enabled: true
    runtimeScanner:
      deploy: true
      settings:
        maxFileSizeAllowed: 2621121212
```

This is the output:

```yaml
# Source: sysdig-deploy/charts/nodeAnalyzer/templates/runtimeScanner/runtime-scanner-configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: sysdig-runtime-scanner
  labels:
    helm.sh/chart: nodeAnalyzer-1.8.50
    app.kubernetes.io/instance: sysdig
    app.kubernetes.io/version: "12.6.0"
    app.kubernetes.io/managed-by: Helm
data:
  api_endpoint: https://secure.sysdig.com
  cluster_name: foo
  debug: "false"
  eve_integration_enabled: "true"
  prom_port: "25001"
  analyzer.maxFileSizeAllowed: "2.621121212e+09"
```

**Workaround**

We can pass in this value as a string, as it is done in the Helm default values.

**Other details**

This chart passes things in through `| quote` but the value is being detected as a floating point number (understandable as that would be the default YAML and JSON behavior, since JSON treats everything as a float64): https://github.com/sysdiglabs/charts/blob/006d2fc36547a0ab8c2f1d6a26bea75ac93651a8/charts/node-analyzer/templates/runtimeScanner/runtime-scanner-configmap.yaml#L37

A possible solution may be to use one of Helm's type conversion functions to force it to be treated as an int64, but I don't know if that causes issues when values are passed in as a string: https://helm.sh/docs/chart_template_guide/function_list/#type-conversion-functions

jawnsy:
Convert maxFileSizeAllowed and maxImageSizeAllowed settings to int64 in Helm, overriding the default behavior (which treats these as float64 values).

This means users can pass either an integer or string, and strings are validated to be int64-compatible.

Closes: #1092

## Checklist


aroberts87:
Hey @jawnsy, thanks for the PR! Would you be able to update your branch before we merge this?

jawnsy:
@aroberts87 I clicked the update branch button here! Should be good to merge now

aroberts87 (approved):

PR #1099, opened 2023-05-15
fix: move away from beta annotations (Close #1098)

### LINKED ISSUE #1098: Using old beta annotations for arch and os
The sysdig charts are using old annotations:

```
Warning: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[1].matchExpressions[0].key: beta.kubernetes.io/arch is deprecated since v1.14; use "kubernetes.io/arch" instead
Warning: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[1].matchExpressions[1].key: beta.kubernetes.io/os is deprecated since v1.14; use "kubernetes.io/os" instead
```

loopingz:
## Checklist

- [ -] Title of the PR starts with type and scope, (e.g. feat(agent,node-analyzer,sysdig-deploy):)
   - Scope is global
  - I would expect semver to handle this
- [X] Variables are documented in the README.md (or README.tpl in some charts) -> N/A

### DIFF
--- a/charts/agent/templates/daemonset.yaml
+++ b/charts/agent/templates/daemonset.yaml
@@ -54,10 +54,10 @@ spec:
             nodeSelectorTerms:
             - matchExpressions:
               {{- if (include "agent.kubeVersionLessThan" (dict "root" . "major" 1 "minor" 14)) }}
-              - key: beta.kubernetes.io/arch
+              - key: kubernetes.io/arch
                 operator: In
                 values: {{- toYaml .Values.daemonset.arch | nindent 18 }}
-              - key: beta.kubernetes.io/os
+              - key: kubernetes.io/os
                 operator: In
                 values: {{- toYaml .Values.daemonset.os | nindent 18 }}
               {{- else }}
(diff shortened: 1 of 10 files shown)


aroberts87:
Hello @loopingz! Thank you for the PR! Can you elaborate a bit more about which charts(s) caused the warnings shown in the linked issue? The reason I ask is because several charts have a check for this and will use the correct annotation depending on cluster version, including a few charts that you've made changes to in this PR.
* [Agent DaemonSet](https://github.com/sysdiglabs/charts/blob/master/charts/agent/templates/daemonset.yaml#L56-L70)
* [Agent Deployment](https://github.com/sysdiglabs/charts/blob/master/charts/agent/templates/deployment.yaml#L39-L53)
* [Rapid Response DaemonSet](https://github.com/sysdiglabs/charts/blob/master/charts/rapid-response/templates/daemonset.yaml#L44-L58) (thought the version compared looks incorrect...)

loopingz:
Ok i ll have another look, we use the templated version to deploy and noticed it was badly generated, but might be condition failing for detection.

loopingz:
So the PR does not make sense as I just run a replacement while you had some conditions in place.

Sorry I should have be more careful on the update

PR #1118, opened 2023-05-22
feat(admission-controller): Deprecate sysdig.url (keep for compatibility) and use apiEndpoint instead

airadier:

As discussed internally, the AC chart  uses sysdig.url instead of sysdig.apiEndpoint and that expects the URL to have an `https://` prefix.

With this change, the AC chart aligns better with the existing subcharts.

Backwards compatibility is kept by still supporting sysdig.url

## Checklist


charts/admission-controller/templates/webhook/configmap.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   labels:
     {{ include "admissionController.webhook.labels" . | nindent 4 }}
 data:
-  SECURE_URL: "{{ include "admissionController.apiEndpoint" . }}"
+  SECURE_URL: "{{ .Values.sysdig.url | default (printf "https://%s" (include "admissionController.apiEndpoint" .)) }}"
```
mavimo, line 9: Are we sure is always HTTPS?

charts/admission-controller/templates/scanner/configmap.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   labels:
     {{ include "admissionController.scanner.labels" . | nindent 4 }}
 data:
-  SECURE_BASE_URL: "{{ include "admissionController.apiEndpoint" . }}"
+  SECURE_BASE_URL: "{{ .Values.sysdig.url | default (printf "https://%s" (include "admissionController.apiEndpoint" .)) }}"
```
mavimo, line 10: Are we sure is always HTTPS?

airadier: It should be HTTPS, at least for production. It is a strong assumption, but we are making the same assumption in other places like:

- https://github.com/sysdiglabs/charts/blob/master/charts/node-analyzer/templates/configmap-host-analyzer.yaml#L12
- https://github.com/sysdiglabs/charts/blob/master/charts/node-analyzer/templates/configmap-host-scanner.yaml#L10
- https://github.com/sysdiglabs/charts/blob/master/charts/rapid-response/templates/configmap.yaml#L8

dimar-apolicy (commented):
@airadier no objections from my side, as long as it approved by @lior-alafi-work we good to go

lior-alafi-work (approved):
cool once merged i'll pull to the kspm ac branch so we can be aligned with that value

PR #1158, opened 2023-06-12
feat(adminssion-controller,agent,node-analyzer,kspm-collector,rapid-response): pull policy secret params uniformity

francesco-furlan:

Add global  pullSecrets (`global.image.pullSecrets`) and pullPolicy (`global.image.pullPolicy`) params in order to keep consistency between the charts.
The user can continue to use local params and also use them to override the global ones.

## Checklist


charts/admission-controller/templates/webhook/deployment.yaml
```diff
@@ -29,9 +29,12 @@ spec:
                   matchLabels:
                     {{- include "admissionController.webhook.selectorLabels" . | nindent 20 }}
               weight: 10
-      {{- with .Values.webhook.imagePullSecrets }}
+      {{- if .Values.global.image.pullSecrets }}
```
AlbertoBarba: Wouldn't make sense to invert `if` and `else if` statements?
This way we can use `global.image.pullSecrets` as a default but give the user the ability to override it specifying the `webhook.imagePullSecrets`.
WDYT @mavimo ?

francesco-furlan: As agreed on Slack I'm going to change the behaviour with your suggestion, so allowing local override. Thanks!

AlbertoBarba (approved):

aroberts87 (approved):
LGTM!

PR #1159, opened 2023-06-12
fix(agent): trim newlines when applying nodeSelectors in delegatedAgentDeployments

aroberts87:
Without the trim there were occasionally a newline inserted between `nodeSelector:` and the content desired to be present, which could result in the delegated agent deployment Deployment object failing to render.

## Checklist


irozzo-1A (approved):
LGTM

mavimo (approved):

AlbertoBarba (approved):

PR #1181, opened 2023-06-23
fix: cluster name should be optional

ProvoK:
in multi mode it can be empty. Application will raise an error if it needs to be a non empty string

### DIFF
--- a/charts/cluster-scanner/templates/deployment.yaml
+++ b/charts/cluster-scanner/templates/deployment.yaml
@@ -98,7 +98,6 @@ spec:
                 name: {{ include "cluster-scanner.fullname" . }}
                 key: cluster_name
                 optional: true
-          - name: SYSDIG_KUBECONFIG_CONTENT
           - name: K8S_ROOT_NAMESPACE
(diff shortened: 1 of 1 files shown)


charts/cluster-scanner/templates/deployment.yaml
```diff
@@ -97,6 +97,8 @@ spec:
               configMapKeyRef:
                 name: {{ include "cluster-scanner.fullname" . }}
                 key: cluster_name
+                optional: true
+          - name: SYSDIG_KUBECONFIG_CONTENT
```
michele-mangili: Is `- name: SYSDIG_KUBECONFIG_CONTENT` really required?

ProvoK: 😮 non sure why it's there, gonna fix soon

michele-mangili (approved):

charts/cluster-scanner/templates/deployment.yaml
ProvoK: fixed

PR #1190, opened 2023-06-26
feat: 0.4.0 plus caches improvements

ProvoK:

as per title


charts/cluster-scanner/templates/_helpers.tpl
```diff
@@ -118,7 +118,7 @@ ise_js_producer_subject: "analysis.sboms"
 Generates config for Redis, if available
 */}}
 {{- define "cluster-scanner.redisCacheConfig" }}
-{{- if eq .Values.imageSbomExtractor.cache.type "redis" }}
+{{- if and (contains "distributed" .Values.imageSbomExtractor.cache.type) (.Values.imageSbomExtractor.cache.redis) }}
```
michele-mangili, line 121: I am fine with it but as an improvement, I guess we should fail if the `cache.type="distributed"` and `cache.redis` is not provided.

charts/cluster-scanner/values.yaml
```diff
@@ -156,15 +156,19 @@ imageSbomExtractor:
   pprofEnabled: false
 
   cache:
-    # The type of cache to use, either `local` or `redis`. When setting `redis`, both Redis and the local cache will be used.
+    # The type of cache to use.
+    # Allowed values are `local`, `distributed` and `distributed,local`.
+    # When specified more than one, the cache precedence will be applied from the left to the right.
```
michele-mangili: I think the precedence is applied from right to left, since you first check whether the cache on the right does have the data.

```suggestion
    # When specified more than one, the cache precedence will be applied from right to left.
```

michele-mangili (approved):

charts/cluster-scanner/templates/_helpers.tpl
ProvoK: I did so to try to be more future proof: a different distributed cache implementation may be developed in the future and might not be strictly redis

michele-mangili (approved):

PR #1204, opened 2023-06-29
feat(admission-controller, sysdig-deploy): Add logic to toggle enable/disable HPA and set replica count based on that value.

maratsal:
Introducing option to enable/disable HPA for admission controller webhook. Based on that switch we are able to set deployment replica count. This s required to avoid Argo CD confusion where application goes out of sync if replica count is changed based on HPA activity - please see details - https://argo-cd.readthedocs.io/en/stable/user-guide/best_practices/#leaving-room-for-imperativeness

## Checklist


charts/admission-controller/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: admission-controller
 description: Sysdig Admission Controller using Sysdig Secure inline image scanner
 type: application
-version: 0.9.0
+version: 0.9.1
```
aroberts87: I think since this change is introducing a new component that is defaulted to enabled, the HPA, we should make this a minor release instead of a patch. This will require you to also do the version bumps in `sysdig-deploy/Chart.yaml` though.
```suggestion
version: 0.10.0
```

maratsal: bumped minor version as it was suggested.

aroberts87:
@sysdiglabs/cloud-native - could someone please review this? 🙏🏻

wideawakening:
i rather put on title `Add logic to toggle enable/disable HPA` , since it's not disabling it.
rest seems ok, and backwards compatible

mavimo (approved):

mavimo:
@maratsal thx for your contribution!

PR #1214, opened 2023-07-03
feat(admission-controller): Add KSPM Admission Controller to helm chart

airadier:
## Checklist


charts/admission-controller/values.yaml
```diff
@@ -71,6 +73,12 @@ sysdig:
   # Overrides the `global.sysdig.secureAPIToken` parameter.
   secureAPIToken: ""
 
+  # **required for KSPM AC**
+  # <br/>Access Key to access Sysdig Secure.
+  # <br/><br/>Either this value or `sysdig.existingAccessKeySecret` is required
+  # Overrides `global.sysdig.accessKey`
+  accessKey: ""
```
AlbertoBarba: Isn't it the same as `secureAPIToken`?

charts/admission-controller/values.yaml
```diff
@@ -12,6 +12,8 @@ global:
     secureAPITokenSecret: ""
     # The global Sysdig Secure region.
     region: "us1"
+    # The global Access Key to access Sysdig Secure.
+    accessKey: ""
```
AlbertoBarba, line 16: Isn't it the same as `secureAPIToken`?

charts/admission-controller/values.yaml
lilx1ao: I don't think so. check: https://sysdig.atlassian.net/wiki/spaces/CS/pages/2945319073/Glia+Admission+Controller+Implementation+Use+Case

airadier: No. AccessKey is the "agent installation access key", while the API token is the user or service account token. There were requirements to use access key instead of API token for cluster components, and that's why the new AC uses access key (like agent and the rest of components)

charts/admission-controller/values.yaml
airadier: See comment above. This one is for the global override, while the one in line 80 is the local one.

charts/admission-controller/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: admission-controller
 description: Sysdig Admission Controller using Sysdig Secure inline image scanner
 type: application
-version: 0.12.2
+version: 0.12.3
```
aroberts87: Could we please at least bump the minor version here since we're adding a whole new component?

charts/sysdig-deploy/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sysdig-deploy
 description: A chart with various Sysdig components for Kubernetes
 type: application
-version: 1.17.6
+version: 1.17.7
```
aroberts87: Similar to my earlier request to bump the minor version due to adding a whole new component.

charts/admission-controller/values.yaml
AlbertoBarba: Since `sysdig-deploy` already supports it, can u add also the `accessKeySecret` and its usage?

charts/admission-controller/README.md
```diff
@@ -110,9 +111,11 @@ The following table lists the configurable parameters of the `admission-controll
...
+| sysdig.accessKey                                   | **required for KSPM AC** <br/>Access Key to access Sysdig Secure. <br/><br/>Either this value or `sysdig.existingAccessKeySecret` is required Overrides `global.sysdig.accessKey`                                                                                                                                                                                                                                                                             | <code>""</code>                                                                                                                                                                                    |
```
aroberts87: nit: Could you please expand AC to Admission Controller like what is done in the description for the `features.kspmAdmissionController` parameter below?

charts/admission-controller/values.yaml
```diff
@@ -152,6 +163,24 @@ podMonitors:
...
+    service:
+      # Use this type as webhook service
+      type: ClusterIP
+      # Configure port for the V2 webhook service
+      port: 7443
+    http:
+      # HTTP serve port where the requests will be served from
+      port: 7443
+      # KSPM Admission Controller container image
+    image: quay.io/sysdig/secure-admission-controller:1.25.5-rc
```
aroberts87: For the other container images we have we separate the registry, repository, and tag values to be independently configurable. Can we adopt that methodology here?

As examples:
- Agent: https://github.com/sysdiglabs/charts/blob/master/charts/agent/values.yaml#L57-L59
- Admission Controller: https://github.com/sysdiglabs/charts/blob/master/charts/admission-controller/values.yaml#L162-L168
- KSPM Collector: https://github.com/sysdiglabs/charts/blob/master/charts/kspm-collector/values.yaml#L77-L82

charts/admission-controller/values.yaml
airadier: That's in our backlog - SSPROD-27044

The change is not completely trivial. If you consider this mandatory for this release (this will be Controlled Availability) I can prioritize that bug ticket. Otherwise, I would release without it and the team will handle the implementation.

charts/admission-controller/Chart.yaml
airadier: You are right, and in fact I bumped the version but then the master branch catch up on 0.12 :) 

Bumping it.

charts/admission-controller/values.yaml
airadier: Created issue for tracking: SSPROD-28668.

Similar to https://github.com/sysdiglabs/charts/pull/1214#discussion_r1287404928, should we block the Controlled Availability release and prioritize the ticket, or can we release CA and address this later?

charts/admission-controller/values.yaml
AlbertoBarba: Just thinking about those users that already have the `accessKeySecret`. If they want to enable this `kspmAdmissionController` they'll be forced to add `accessKey` as well 😅

charts/admission-controller/values.yaml
AlbertoBarba: @airadier I guess so...otherwise we'll create a BC

charts/admission-controller/values.yaml
airadier: You are right, I'll try to expedite this

charts/admission-controller/values.yaml
airadier: This should be addressed now.

charts/admission-controller/values.yaml
airadier: This should be addressed now too, but I am waiting for a minor change in the AC image to read the config from `access-key` instead of `access_key`.

The current key convention for the existingAcessKeySecret is using the `access-key` as the key name inside the secret. Now I had to add a projected volume to remap to the proper path. Once this is fixed I'll be able to simplify.

charts/admission-controller/templates/webhook/deployment.yaml
```diff
@@ -109,6 +118,8 @@ spec:
             {{- end }}
           imagePullPolicy: {{ .Values.webhook.image.pullPolicy | default .Values.global.image.pullPolicy }}
           env:
+            - name: foo
```
airadier: This was introduced by mistake, removing in next commit

AlbertoBarba (approved):
LGTM

PR #1217, opened 2023-07-05
fix(admission-controller): Make ValidatingWebhookConfiguration part of the resource lifecycle

airadier:

Currently, the ValidatingWebhookConfiguration is created using a post-upgrade and post-install hook. This was made to prevent the webhook registration blocking the deployment of the AC components itself, so it is not registered until AC is correctly deployed. But, **this means the resource is not managed as part of the Helm Chart resource lifecycle, and it won't be removed on uninstall**.

More info: https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases

Keeping the webhook registration while the services are not available might mean that the cluster is degraded because there is a configured AC endpoint but the service won't be available.

The fix is creating the  ValidatingWebhookConfiguration as a standard resource, but with an empty webhook list to prevent it from blocking AC components. The post-upgrade and post-install hooks will apply the same resource again but with the right webhooks depending on the configuration.

As the ValidatingWebhookConfiguration is now part of the lifecycle, it will be properly removed on uninstall.

## Checklist

### DIFF
--- a/charts/admission-controller/Chart.yaml
+++ b/charts/admission-controller/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: admission-controller
 description: Sysdig Admission Controller using Sysdig Secure inline image scanner
 type: application
-version: 0.10.0
+version: 0.10.1
 appVersion: 3.9.22
 home: https://sysdiglabs.github.io/admission-controller/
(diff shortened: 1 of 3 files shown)


miketnt (approved):
LGTM 👍🏼

PR #1246, opened 2023-07-20
feat(admission-controller): simplify CA and certificate options

airadier:

As discussed internally, there are no known requests to provide custom certificates or CA for the internal webhook, as it is automatically generated and the right CA bundle is provided on registration.

Moreover, the current implementation mixes CA certificate for trust with internal TLS certificate generation, which is not right. CA key should not be provided to the Helm chart

## Checklist


charts/admission-controller/values.yaml
```diff
@@ -220,26 +220,8 @@ webhook:
...
-      # For outbound connections (secure backend, proxy,...)
-      # <br/>A PEM-encoded private key of the certificate authority to use in the certificate generation.
-      # <br/>If empty, a new CA will be autogenerated.
-      key: ""
-    # For inbound connections to serve HttpRequests as Kubernetes Webhook.
-    # <br/>A PEM-encoded x509 certificate signed by the CA.
-    # <br/>If empty, a new cert will be generated.
-    # <br/>If provided, it must be valid with the `webhook.ssl.ca`.
-    # <br/>If this is set, the key must also be provided.
-    cert: ""
```
AlbertoBarba, line 236: We can probably keep these `cert` inside values since they r going to be used in https://github.com/sysdiglabs/charts/pull/1246/files#diff-154ad88a2fe7629ec86703c63fa030d26afc204bd238ff8a662709298bfc69beL178

charts/admission-controller/values.yaml
```diff
@@ -220,26 +220,8 @@ webhook:
...
-    # For inbound connections to serve HttpRequests as Kubernetes Webhook.
-    # <br/>A PEM-encoded private key signed by the CA.
-    # <br/>If empty, a new key will be generated.
-    # <br/>If provided, it must be valid with the `webhook.ssl.ca`.
-    # <br/>If this is set, the cert must also be provided.
-    key: ""
```
AlbertoBarba, line 242: We can probably keep these `key` inside values since they r going to be used in https://github.com/sysdiglabs/charts/pull/1246/files#diff-154ad88a2fe7629ec86703c63fa030d26afc204bd238ff8a662709298bfc69beL178

AlbertoBarba (approved):

PR #1259, opened 2023-07-24
feat(agent,rapid-response): set metadata.namespace on all namespaced items

aroberts87:
There were a handful of places in the charts related to `sysdig-deploy` where the `metadata.namespace` field was not being explicitly set. This is works out fine as Helm will set those fields itself during the `helm install` process, but does not do so during a call to `helm template`. This has been found to cause issues with come CI platforms that  generate manifest files via `helm template` and then apply them later with `kubectl apply -f ...`. What has been observed is that these select CI platforms will patch the manifests of namespaced items that do not have `metadata.namespace` set, and will do so with an unexpected value. When the subsequent `kubectl apply -f ...` command is run that supplies the desired namespace, the command fails because the namespace field is already set on these items and is different from what is being requested. This change ensures that all namespaced items have the values of `metadata.namespace` explicitly set to prevent the above issue.

## Checklist

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -5,7 +5,7 @@ description: Sysdig Monitor and Secure agent
 type: application
 
 # currently matching sysdig 1.14.32
-version: 1.11.1
+version: 1.12.0
 
 appVersion: 12.15.0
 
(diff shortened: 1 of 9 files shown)


dark-vex (approved):
Thanks @aroberts87 LGTM

mavimo (approved):

PR #1260, opened 2023-07-24
feat(doc): Move On-prem deployment instruction to the registry scanner chart

radhikapc:
## Checklist


charts/registry-scanner/README.tpl
```diff
@@ -50,3 +50,23 @@ The following table lists the configurable parameters of the Sysdig Registry Sca
 
 {{ .Chart.Values }}
 {{- end -}}
+
+## On-Prem Deployment
+
+If you are using a Sysdig on-prem version greater than 6.2, you need to configure legacy VM engine while setting up registry scanner.
```
airadier, line 56: Maybe I would add this as a side note instead? It is not clear if this won't work for <6.2, but I understand legacy VM engine is enabled by default? The note at the top makes me think that the following instructions might not work if version < 6.2

wideawakening: my 2 cents for clarification; this helm snippet works both for < 6.2 and > 6.2.
for lower than, legacy VM services are available by default, nothing to be done
for upper than, however, legacy VM services are only available by default, must be enabled  (which i don´t know what feature flag is used), otherwise it would not work

radhikapc: Do you mean:  if you are on a version greater than v6.2, you need to explicitly enable legacy VM service, and on versions prior to v6.2, the legacy engine is enabled by default.

radhikapc: Updated the text

airadier (approved):

PR #1264, opened 2023-07-25
feat(doc): Revamp the  landing page of Sysdig Helm Chart

radhikapc:

DOC-3252

## Checklist


mavimo (approved):

index.md
```diff
@@ -1,47 +1,42 @@
...
+|                                                | [`node-analyzer`](/charts/node-analyzer)               | The `node-analyzer` deploys the following as one daemonset in a cluster: <br>`host-scanner`<br>`runtime-scanner`<br>`kspm-analyzer` |
+|                                                | [`kspm-collector`](/charts/kspm-collector)             | Collects Kubernetes resource manifests and sends them to be evaluated against Sysdig compliance policies. The scan results are displayed in Sysdig Secure [Compliance](https://docs.sysdig.com/en/compliance/) and [Inventory ](https://docs.sysdig.com/en/docs/sysdig-secure/inventory/)UI. |
+|                                                | [`host-scanner`](/charts/node-analyzer)                | Scan [hosts](https://docs.sysdig.com/en/docs/sysdig-secure/vulnerabilities/runtime/host-scanning/) to identify and prioritize vulnerabilities with rich context. |
+|                                                | [`runtime-scanner`](/charts/node-analyzer)             | Scan Kubernetes [workloads](https://docs.sysdig.com/en/docs/sysdig-secure/vulnerabilities/runtime/) in runtime to identify and prioritize vulnerabilities with rich context. |
```
aroberts87: Having these called out in the `Subcharts` column is a bit misleading as they are not their own charts, but rather a part of the `node-analyzer` chart. It's possible someone would get confused clicking on the `host-scanner` link and immediately landing on a page called `Node Analyzer`. 🤷🏻

index.md
```diff
@@ -1,47 +1,42 @@
...
-|[`sysdig-mcm-navmenu`](/charts/sysdig-mcm-navmenu) | Deploys services and ingress required to create Multi-Cloud Management menu in Sysdig. |
-|[`sysdig-stackdriver-bridge`](/charts/sysdig-stackdriver-bridge) | Deploys the the Sysdig Stackdriver Bridge application to enable reading and sending audit logs from Stackdriver to Sysdig agent. |
+Additional Sysdig Secure components provide scanning for vulnerabilities in various parts of the environment and the CI/CD lifecycle.
 
+| Parent Chart                                   | Subcharts                                              | Description                                                  |
+| ---------------------------------------------- | :----------------------------------------------------- | ------------------------------------------------------------ |
+| [`sysdig-deploy`](/charts/sysdig-deploy)       |                                                        | Installs, configures, and upgrades the Sysdig Agent and a variety of Sysdig Secure components. The parent chart deploys multiple child charts. |
```
aroberts87, line 20: It looks like the `cluster-scanner` subchart has been missed here. It was added to `sysdig-deploy` a few weeks back and can be found [here](https://github.com/sysdiglabs/charts/tree/master/charts/cluster-scanner). I know the feature is in CA, so maybe this omission was intentional, but it's already included in the chart so I see no problem with documenting it here. @mavimo , do you have thoughts on this?

index.md
amy-lodato: Good point. I wonder if it's still worthwhile to call out the components but begin the description with "Component deployed in node-analyzer chart to...[scan hosts etc]" and ""Component deployed in node-analyzer chart to...[scan Kubernetes workloads etc]"

index.md
amy-lodato: Need advice from @mavimo -- till now we were told not to mention it

mavimo: @aroberts87 I personally suggest waiting until it goes GA. I know this is available in the codebase but this is more of a "documentation" page and perhaps it is worth listing the charts we support; if we want to add it IMHO we should state that it is in CA.

FYI JM2C, I'm not concerned too much on adding the `cluster-scanner` as well.

aroberts87: Sounds good to me! Let's 🚢 it! 🙂

aroberts87 (approved):

PR #1270, opened 2023-07-27
feature(agent): add support for Universal eBPF

mavimo:

 - Add support for the Universal eBPF driver

## Checklist


charts/agent/templates/daemonset.yaml
```diff
@@ -95,6 +96,13 @@ spec:
             - name: SYSDIG_BPF_PROBE
               value:
           {{- end }}
+          {{- if (include "agent.universalEbpfEnforced" .) }}
```
AlbertoBarba: I think this env is not needed...if `agent.universalEbpfEnforced` returns true we don't start the `initContainer`

AlbertoBarba (changes requested):

AlbertoBarba (approved):

charts/agent/tests/universal_ebpf_test.yaml
```diff
@@ -0,0 +1,103 @@
+suite: Universal eBPF tests
+templates:
+  - templates/daemonset.yaml
+tests:
+  - it: Ensure that when the eBPF is disabled we create the sysdig container wihtout SYSDIG_BPF_PROBE and SYSDIG_AGENT_DRIVER environment variables
```
aroberts87:
```suggestion
  - it: Ensure that when the eBPF is disabled we create the sysdig container without SYSDIG_BPF_PROBE and SYSDIG_AGENT_DRIVER environment variables
```

charts/agent/values.yaml
```diff
@@ -169,6 +169,10 @@ priorityClassValue: 10
 ebpf:
   # Enable eBPF support for Sysdig Agent
   enabled: false
+
+  # Define the kind of eBPF driver that can be used in the agent. Can be `legacy_ebpf` or `universal_ebpf`
```
aroberts87:
```suggestion
  # Define the kind of eBPF driver that will be used by the agent. Can be `legacy_ebpf` or `universal_ebpf`
```

aroberts87 (approved):
LGTM! Just two small nits.

PR #1272, opened 2023-07-28
fix(node-analyzer): permissions fix for CR sysdig-agent-node-analyzer

### LINKED ISSUE #1253: Node Analyzer Benchmark Write Permissions Toggle Statements Duplicated
In the Benchmark Runner Cluster Role there's a toggle based on `.Values.nodeAnalyzer.benchmarkRunner.includeSensitivePermissions` that should be stripping out write permissions required in the role. Above that there is a static statement that gives the Cluster Role for Node Analyzer the permissions regardless. If the value is set to true the same clause shows up twice in the role definition. 

This is the only thing that requires cluster wide pod creation and deletion permissions so if it's not needed I would like to strip these permissions out. If they are needed for Node Analyzer's standard operation, take the toggle out to avoid confusion about if the permissions are necessary.

```yaml
# https://github.com/sysdiglabs/charts/blob/master/charts/node-analyzer/templates/clusterrole-node-analyzer.yaml#L36
- apiGroups:
    - ""
  resources:
    - "pods"
    - "pods/exec"
  verbs:
    - "create"
- apiGroups:
    - ""
  resources:
    - "pods"
  verbs:
    - "delete"

# https://github.com/sysdiglabs/charts/blob/master/charts/node-analyzer/templates/clusterrole-node-analyzer.yaml#L99
{{- if .Values.nodeAnalyzer.benchmarkRunner.includeSensitivePermissions  }}
- apiGroups:
    - ""
  resources:
    - pods/exec
  verbs:
    - create
- apiGroups:
    - ""
  resources:
    - pods
  verbs:
    - create
    - delete
{{- end }}
```

a1ex-var1amov:

The PR removes extensive permissions that were provided to ClusterRole sysdig-agent-node-analyzer by the following [commit](https://github.com/sysdiglabs/charts/commit/dedebca2a44e00828d80379c00cafbb50d90db8d). Per tests, the permissions were required for the BenchMark runner; however, kspm-analyzer doesn't require them to function properly. 

The PR is the result of investigating the issue [1253](https://github.com/sysdiglabs/charts/issues/1253).

Please double-check. Thank you!

## Checklist

- [n/a] Variables are documented in the README.md (or README.tpl in some charts)
- [n/a] All test files are added in the tests folder of their respective chart and have a "_test" suffix


ProvoK (approved):
LGTM for what concerns NIA/RuntimeScanner/HostScanner

charts/node-analyzer/templates/clusterrole-node-analyzer.yaml
```diff
@@ -33,19 +33,6 @@ rules:
     - "flowschemas"
   verbs:
     - list
-- apiGroups:
```
dark-vex, line 36: @a1ex-var1amov we still have on-premise customers using the Benchmark Runner, can we add `{{- if (include "nodeAnalyzer.deployBenchmarkRunner" .) }}` for add those additional permission only when `benchmarkRunner.deploy` is set to `true` ?

dark-vex (changes requested):

chen-shmilovich-sysdig (approved):

charts/node-analyzer/templates/clusterrole-node-analyzer.yaml
dark-vex: Thanks Alex for fixing it

dark-vex (approved):

PR #1273, opened 2023-07-31
chore(node-analyzer): toggle volumes used by runtimeScanner and imageAnalyzer

francesco-furlan:

Don't mount volumes needed just for runtimeScanner and imageAnalyzer when their both disabled.
This is needed since GKE autopilot doesn’t allow mounting volumes not used.
Preliminary work for SMAGENT-5237 and SMAGENT-5236

## Checklist

### DIFF
--- a/charts/node-analyzer/templates/daemonset-node-analyzer.yaml
+++ b/charts/node-analyzer/templates/daemonset-node-analyzer.yaml
@@ -40,7 +40,7 @@ spec:
             name: {{ include "nodeAnalyzer.configmapName" . }}
             optional: true
         {{- end }}
-        {{- if include "nodeAnalyzer.deployRuntimeScanner" . }}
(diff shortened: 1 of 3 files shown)


charts/node-analyzer/templates/daemonset-node-analyzer.yaml
```diff
@@ -40,6 +40,7 @@ spec:
             name: {{ include "agent.configmapName" . }}
             optional: true
         {{- end }}
+        {{- if include "nodeAnalyzer.deployRuntimeScanner" . }}
```
AlbertoBarba: It seems some of the volumes defined here (like `vardata-vol` and `varrun-vol`) are used also in other services, such as the `imageAnalyzer`.

francesco-furlan: Nice catch! 👍

AlbertoBarba (approved):

PR #1287, opened 2023-08-07
feat(cluster-scanner): Introduce support to external secrets accessKeySecret

dark-vex:
Introduce the possibility to use external secret for cluster scanner.
Currently it is possible to specify a `global.sysdig.accessKeySecret` for all the components besides the cluster scanner.

When using the external secret `accessKeySecret` the environment variable `SYSDIG_ACCESS_KEY` will look for the key `access-key` instead of `sysdig_access_key`, I did that in order to align with the other charts

## Checklist


charts/cluster-scanner/templates/secret.yaml
```diff
@@ -1,3 +1,18 @@
+{{- if (include "cluster-scanner.accessKeySecret" .) }}
```
AlbertoBarba: Instead of having this huge `if` statement, wouldn't it be better to have it arount the `sysdig_access_key` key (https://github.com/sysdiglabs/charts/pull/1287/files#diff-20cae487430fbff57280c8c521a4ed366a365219b54761f1dab3165a8583ee5dR25)?

AlbertoBarba (changes requested):

charts/cluster-scanner/templates/secret.yaml
dark-vex: oh yeah, would be definitely better, let me change it

dark-vex: I have changed it, let me know if there is anything else 💪

charts/cluster-scanner/tests/secret_test.yaml
```diff
@@ -60,3 +60,34 @@ tests:
     asserts:
       - isNull:
           path: data.cache_redis_password
+
+---
```
AlbertoBarba: I just tried locally and it seems these new tests are not actually running. I just tried to change input and expectations in order to see it fail and it doesn't 😕

AlbertoBarba (changes requested):

AlbertoBarba (approved):
LGTM 🚀

PR #1302, opened 2023-08-22
feat(cluster-scanner): allow to configure http proxy

AlbertoBarba:

Cluster scanner currently had the possibility to specify a custom CA but it was not adding the needed variables needed for HTTP/HTTPS proxy.

## Checklist

### DIFF
--- a/charts/cluster-scanner/Chart.yaml
+++ b/charts/cluster-scanner/Chart.yaml
@@ -4,7 +4,7 @@ description: Sysdig Cluster Scanner
 
 type: application
 
-version: 0.4.1
+version: 0.5.0
 
 appVersion: "0.1.0"
 home: https://www.sysdig.com/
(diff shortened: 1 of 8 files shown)


charts/cluster-scanner/tests/deployment_test.yaml
```diff
@@ -92,9 +92,9 @@ tests:
     asserts:
       - not: true
         isEmpty:
-          path: spec.template.spec.containers[0].env[9]
+          path: spec.template.spec.containers[0].env[12]
```
mavimo: Please use a better way to identify the env node, eg:
https://github.com/sysdiglabs/charts/blob/master/charts/agent/tests/conditional_flag_test.yaml#L17
(same for the other nodes here)

mavimo (changes requested):

charts/cluster-scanner/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: Sysdig Cluster Scanner
 
 type: application
 
-version: 0.4.1
+version: 0.4.2
```
mavimo: I'll bump as minor

mavimo (changes requested):

mavimo (approved):

Issue #1310, opened 2023-08-24
[admission-controller] Helm chart generates duplicate ValidatingWebhookConfiguration resources


omaen:
Hi,

Previously we used terraform to deploy the `sysdig-deploy` helm chart to our cluster, but are now trying to move the setup to ArgoCD using Kustomize. This works for all our other charts, but it seems like the `admission-controller` sub-chart is generating multiple resources with the same name/id, which causes kustomization to fail with the following message
```
Error: may not add resource with an already registered id: admissionregistration.k8s.io_v1_ValidatingWebhookConfiguration|sysdig-agent|sysdig-cluster-agent-admissioncontroller-webhook
```

Our kustomization.yaml, and the `values.yaml` has set `admissionController.enabled: true`:
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
helmCharts:
  - name: sysdig-deploy
    namespace: sysdig-agent
    repo: https://charts.sysdig.com/
    releaseName: sysdig-cluster-agent
    valuesFile: values.yaml
    version: "1.19.2"
    includeCRDs: true
```

Running `helm template sysdig sysdig/sysdig-deploy --values values.yaml` works, but I can see the two resources created.
```
# Source: sysdig-deploy/charts/admissionController/templates/webhook/admissionregistration.yaml
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingWebhookConfiguration
metadata:
  name: sysdig-admissioncontroller-webhook
  namespace: default
webhooks: []
```
and
```
# Source: sysdig-deploy/charts/admissionController/templates/webhook/admissionregistration.yaml
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingWebhookConfiguration
metadata:
  name: sysdig-admissioncontroller-webhook
  namespace: default
  annotations:
    "helm.sh/hook": "post-install, post-upgrade"
    meta.helm.sh/release-name: sysdig
    meta.helm.sh/release-namespace: default
  labels:
    app.kubernetes.io/managed-by: Helm
webhooks:
- name: audit.secure.sysdig.com
  matchPolicy: Equivalent
  rules:
    <truncated>
  clientConfig:
    <truncated>
  admissionReviewVersions: ["v1", "v1beta1"]
  sideEffects: None
  timeoutSeconds: 5
  failurePolicy: Ignore
```

[...]

PR #1313, opened 2023-08-28
fix(admission-controller): Use namespaceSelector to prevent webhook deadlock

omaen:
As mentioned in #1310 the helm-chart for admission-controller produces multiple values for the ValidatingWebhookConfiguration. I was explained that this is because the resource is added with a post-install, and in order to have to resource a part of the lifecycle of the helm-installation the resource needs to be created on install.
By excluding the admission-controller from the webhook validation the issue with deadlocks on installation should be resolved.

I can't see where and how the post-install configuration is done. Can anyone help with this?

## Checklist

### DIFF
--- a/charts/admission-controller/templates/_helpers.tpl
+++ b/charts/admission-controller/templates/_helpers.tpl
@@ -143,7 +143,7 @@ matchExpressions:
   - key: kubernetes.io/metadata.name
     operator: NotIn
     values:
-      - {{ .Release.Namespace }}
+      - {{ .admissionController.namespace }}
 {{- end -}}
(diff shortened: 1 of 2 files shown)


airadier:
Hi @omaen , I just noticed your PR after pushing mine in #1316 . Managing the change in #1316 as I address backwards compatibility with previous charts version that already deployed using the post-install hook.

omaen:
Hey @airadier, no worries. It's my first time contributing here, and I had some issues with generating new docs with the pre-commit script. Thanks for the help in resolving the issue :)

PR #1315, opened 2023-08-28
fix(admission-controller): Use EKS Terraform friendly ports

airadier:

EKS clusters, when deployed via official TF module https://registry.terraform.io/modules/terraform-aws-modules/eks/aws/latest, get a strict security group that blocks communication from the kubeapi to the AC webhooks.

This security group includes some rules meant to allow this communication to webhooks:
- 4443
- 6443
- 8443
- 9443

This PR changes the pod ports to 6443 and 8443 which are allowed by default in the cluster SG.

CC @maratsal 

## Checklist


AlbertoBarba (commented):
Is it really worth changing the default values when the customers can change them on their own?
I think it might be a possible BC for customers who currently use default values and have security groups (or equivalent) in line with the current configuration 🤔

airadier:
> Is it really worth changing the default values when the customers can change them on their own?
> I think it might be a possible BC for customers who currently use default values and have security groups (or equivalent) in line with the current configuration 🤔

Yeah agree, and I thought about it when pushing the changes. I am not aware how many customers are facing this issue (maybe @maratsal can provide some context). 

As a middle ground, I would propose changing only the defaults for the new KSPM Admission Controller, which is Controlled Availability and not actively in use by customers, while keeping the defaults for the previous webhook unchanged. What do you think?

maratsal:
I am not aware of customer cases myself too, I noticed this EKS Terraform module particularity in my lab environment where I realized if Sysdig Admission Controller is installed on EKS environment deployed by official EKS terraform module, then we need add additional security group rules. Here is example with working EKS terraform module:
```
module "eks" {
  source  = "terraform-aws-modules/eks/aws"
  version = "19.14.0"

  cluster_name    = local.cluster_name
  cluster_version = var.eks_cluster_version

  vpc_id                         = module.vpc.vpc_id
  subnet_ids                     = module.vpc.private_subnets
  cluster_endpoint_public_access = true

  eks_managed_node_group_defaults = {
    ami_type = "AL2_x86_64"
  }

  eks_managed_node_groups = var.eks_managed_node_groups

  # additional node security group rules to allow access from control plane to Sysdig Admission Controller webhook port
  node_security_group_additional_rules = {
    ingress_allow_access_from_control_plane_to_sysdig_ac = {
      type                          = "ingress"
      protocol                      = "tcp"
      from_port                     = 5000
      to_port                       = 5000
      source_cluster_security_group = true
      description                   = "Allow access from control plane to webhook port of Sysdig Admission Controller"
    }
    ingress_allow_access_from_control_plane_to_sysdig_kspm_ac = {
      type                          = "ingress"
      protocol                      = "tcp"
      from_port                     = 7443
      to_port                       = 7443
      source_cluster_security_group = true
      description                   = "Allow access from control plane to webhook port of Sysdig KSPM Admission Controller"
    }
  }
}

```

AlbertoBarba:
> I am not aware of customer cases myself too, I noticed this EKS Terraform module particularity in my lab environment where I realized if Sysdig Admission Controller is installed on EKS environment deployed by official EKS terraform module, then we need add additional security group rules. Here is example with working EKS terraform module:
> 
> ```
> module "eks" {
>   source  = "terraform-aws-modules/eks/aws"
>   version = "19.14.0"
> 
>   cluster_name    = local.cluster_name
>   cluster_version = var.eks_cluster_version
> 
>   vpc_id                         = module.vpc.vpc_id
>   subnet_ids                     = module.vpc.private_subnets
>   cluster_endpoint_public_access = true
> 
>   eks_managed_node_group_defaults = {
>     ami_type = "AL2_x86_64"
>   }
> 
>   eks_managed_node_groups = var.eks_managed_node_groups
> 
>   # additional node security group rules to allow access from control plane to Sysdig Admission Controller webhook port
>   node_security_group_additional_rules = {
>     ingress_allow_access_from_control_plane_to_sysdig_ac = {
>       type                          = "ingress"
>       protocol                      = "tcp"
>       from_port                     = 5000
>       to_port                       = 5000
>       source_cluster_security_group = true
>       description                   = "Allow access from control plane to webhook port of Sysdig Admission Controller"
>     }
>     ingress_allow_access_from_control_plane_to_sysdig_kspm_ac = {
>       type                          = "ingress"
>       protocol                      = "tcp"
>       from_port                     = 7443
>       to_port                       = 7443
>       source_cluster_security_group = true
>       description                   = "Allow access from control plane to webhook port of Sysdig KSPM Admission Controller"
>     }
>   }
> }
> ```

@maratsal Another solution might be to deploy the helm chart with additional parameters (`--set webhook.http.port=8443,webhook.v2.http.port=6443,webhook.v2.service.port=6443`). Maybe, instead of changing the default values, we can ask to improve the documentation about this topic. WDYT?

airadier:
@maratsal @AlbertoBarba I agree this might be addressed in the documentation, although it seems to be a corner case when using the Terraform EKS module. 

My proposal is that we keep it unchanged for the existing Admission Controller, as customers might be already adding exceptions to their security groups and we would break backwards compatibility. But we use a safer default (port 6443) for new the KSPM Admission Controller, which is under CA and without public documentation yet. This should make it less prone to fail with default security group settings.

As there is no one currently using the KSPM Admission Controller due to the controller availability, I would suggest making the change now so new customers will use the safer default port.

Do you agree? Thanks!

maratsal:
> @maratsal Another solution might be to deploy the helm chart with additional parameters (`--set webhook.http.port=8443,webhook.v2.http.port=6443,webhook.v2.service.port=6443`). Maybe, instead of changing the default values, we can ask to improve the documentation about this topic. WDYT?

@AlbertoBarba yes, that might be another good option. i.e. to add note to documentation page to inform users if they use EKS terraform module to deploy EKS cluster - they might want to change default AC default ports or add additional security group rules.

maratsal:
@airadier yes, I like the approach to change default port for new KSPM AC and keep old one with no change. Also I would still add a note to documentation. Do you want me to create PR for it?

airadier:
> @airadier yes, I like the approach to change default port for new KSPM AC and keep old one with no change. Also I would still add a note to documentation. Do you want me to create PR for it?

Please go ahead. The documentation still does not include references to the new KSPM Admission Controller, but the issue impacts the existing AC too, so feel free to add a note there.

Thanks very much!

AlbertoBarba (approved):

PR #1316, opened 2023-08-29
feat(admission-controller): use namespaceSelector for webhook

airadier:

Fix for #1310 by using namespaceSelector instead of creating an empty VAC and then using post-install hook.

The approach keeps the post-hook for backwards compatibility, so it upgrades any existing post-hook to include the annotations so it is managed by the helm chart lifecycle.

When generating a template, post-install hook should be ignored, fixing the issue with duplicate resource.

Duplicates #1313 

Co-authored-by: @omaen

## Checklist


charts/admission-controller/templates/webhook/admissionregistration.yaml
```diff
@@ -4,115 +4,27 @@ so the template is executed just once
...
-{{- end}}
----
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   name: {{ include "admissionController.webhook.fullname" . }}
   namespace: {{ include "admissionController.namespace" . }}
+{{- if (or (not $existingVac) (and $existingVac (eq $existingVacMetadataAnnotationsReleaseName .Release.Name ) (eq $existingVacMetadataAnnotationsReleaseNamespace .Release.Namespace ))) }}
+{{- include "admissionController.webhookTemplate"  (merge (dict "Cert" $certList._2) . ) }}
+{{- else }}
```
airadier, line 20: We prevent duplicate resource as described in #1310 by either using post-install hook (if we detect the VAC already exists but doesn't have the proper helm lifecycle annotations), or using a standard resource without hook if the VAC doesn't exist, or it exists and has the proper annotations (clean installs, or resource already annotated)

charts/admission-controller/templates/_helpers.tpl
```diff
@@ -376,3 +376,109 @@ an error if not.
 true
 {{- end }}
 {{- end }}
+
+{{- define "admissionController.webhookTemplate" }}
+webhooks:
+{{- if .Values.features.kspmAdmissionController}}
+- name: vac.secure.sysdig.com
+  namespaceSelector:
```
airadier, line 384: Adding the namespaceSelector to prevent issues with the AC blocking itself, instead of the post-install hook approach. From k8s changelog, this will only work in Kubernetes 1.21 or higher, where the kubernetes.io/metadata.name was added. But this should be old enough.

mavimo (approved):

hayk99 (approved):

miketnt (approved):

PR #1337, opened 2023-09-05
docs(registry-scanner): remove nodeSelector from docs

miketnt:
## Checklist


airadier (changes requested):
Sorry but I disagree with the solution, not with the implementation.

The original issue was raised due to a confusion in the usage of `nodeSelector`. Customer was expecting the nodeSelector to apply to the worker jobs, and not only to the orchestrator.

But in current implementation, the nodeSelector applies to the "orchestrator" pod only.

The docs are not clear about this:
"Configure `nodeSelector` for scheduling the registry scanner pod".

I think the solution is to clarify the misunderstanding with the customer.

Then, if this is a requirement, we will need a Feature Request to implement correctly, as we did for the `customLabels` that now also apply to the worker jobs (the issue here was an Admission Controller with mandatory labels).

If this is not a requirement, we can just leave it as is, we can improve the documentation, or we can remove it from values.yaml and the doc. But in this case, we need to consider other parameters in the exact same situation, at least:

- tolerations
- affinity
- podSecurityContext

labels and securityContext where previously addressed as part of a previous Feature Request, and now it is being properly propagated to the worker jobs.

airadier (approved):

PR #1354, opened 2023-09-12
feat(registry-scanner): add gar, gcr and nexus support

maratsal:
## Checklist


charts/registry-scanner/templates/secret.yaml
```diff
@@ -14,10 +14,10 @@ data:
   aws_region: {{ required "A valid .Values.config.aws.region is required" .Values.config.aws.region | b64enc | quote }}
   {{- else if eq .Values.config.registryType "gcr" }}
   registryUser: {{ "_json_key" | b64enc | quote }}
-  registryPassword: {{ required "A valid .Values.config.registryPassword is required" (include "registry-scanner.sanitizeJSON" (.Values.config.registryPassword | toJson)) | b64enc |quote }}
+  registryPassword: {{ required "A valid .Values.config.registryPassword is required" .Values.config.registryUser | b64enc | quote }}
   {{- else if eq .Values.config.registryType "gar" }}
   registryUser: {{ "_json_key_base64" | b64enc | quote }}
```
maratsal, line 16: btw, GAR registry supports both user types: _json_key and _json_key_base64. For _json_key we can provide not encoded json key and for _json_key_base64 we need to provide encoded one.

that's why I haven't touched this logic to let customer decide on what they want.

jprieto92: @airadier what do you think about this? I think that is more reasonable to just support `_json_key_base64` on GAR to have an easy way to put data into the set command as we seen last week.

balinaveen:
@sysdiglabs/vm-integrations  can you please take a look at this PR and provide feedback? @airadier  probably need your final thumbs up too.

charts/registry-scanner/templates/secret.yaml
airadier: Agree, let's keep it. The problem is with the --set, but you can provide a sanitized JSON key via a values.yaml file (tested)

charts/registry-scanner/templates/secret.yaml
```diff
@@ -12,6 +12,9 @@ data:
   aws_access_key_id: {{ .Values.config.aws.accessKeyId | b64enc | quote }}
   aws_secret_access_key: {{ .Values.config.aws.secretAccessKey | b64enc | quote }}
   aws_region: {{ required "A valid .Values.config.aws.region is required" .Values.config.aws.region | b64enc | quote }}
```
airadier, line 14: What's happened with the "gcr" type check here? I still see the code in the comment https://github.com/sysdiglabs/charts/pull/1354#discussion_r1340287484 but not in the final version. I think we should keep it, as we list gcr in the available registry types

airadier (changes requested):

charts/registry-scanner/templates/secret.yaml
jprieto92: For the moment, we have GCR on docs PR as not officially supported. We have 2 options:
- Keep the code as we have previously just o give the possibility to show GCR well on UI (but not officially supported)
- Keep the code as on the final version without a mention to GCR. User can use it as previous to this FR but at least for me, no makes sense to have it when google cloud not give us the possibility to create a new GCR on our accounts because is deprecated and will be removed on google cloud on a few months.

What do you prefer @airadier ?

airadier: @jprieto92 I would keep any code that helps supporting GCR (so the branch in the template that automatically populates the user helps), but no mention either in the docs or in the values.yaml. We might add it in a few days, as we are getting some requests from customers, so it would be just a matter of updating the documentation, but let's keep the code and template as "ready for GCP" as possible. Deal?

jprieto92: after take a look together @airadier latest change here ready to review

jprieto92: done, resolving conversation @airadier

airadier (approved):

PR #1366, opened 2023-09-20
feat(ClusterScanner): NATS JetStream TLS communication

FedericoFeresini:
## Checklist


charts/cluster-scanner/templates/deployment.yaml
```diff
@@ -48,6 +48,20 @@ spec:
           configMap:
             name: {{ include "sysdig.custom_ca.existingConfigMap" (dict "global" .Values.global.ssl "component" .Values.ssl) }}
         {{- end }}
+        {{- if .Values.runtimeStatusIntegrator.natsJS.tls.enabled -}}
+        {{ if eq (include "cluster-scanner.nats.tls.customTLS" .) "false" }}
```
michele-mangili, line 52:
```suggestion
        {{- if eq (include "cluster-scanner.nats.tls.customTLS" .) "false" -}}
```

charts/sysdig-deploy/Chart.yaml
```diff
@@ -42,7 +42,7 @@ dependencies:
   - name: cluster-scanner
     # repository: https://charts.sysdig.com
     repository: file://../cluster-scanner
-    version: ~0.6.0
+    version: ~0.6.1
```
michele-mangili, line 45: I'd recommend you to bump the minor version instead of the patch. See https://github.com/sysdiglabs/charts/pull/1372#discussion_r1336949416

charts/sysdig-deploy/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: sysdig-deploy
 description: A chart with various Sysdig components for Kubernetes
 type: application
-version: 1.23.2
+version: 1.23.3
```
michele-mangili, line 5: I'd bump the chart minor version for the same reason as my comment below.

charts/cluster-scanner/tests/existing-secret_test.yaml
```diff
@@ -21,7 +21,7 @@ tests:
       global.sysdig.accessKeySecret: "secret"
     asserts:
       - equal:
-          path: spec.template.spec.containers[0].env[35]
+          path: spec.template.spec.containers[0].env[39]
```
michele-mangili, line 24: Instead of using the positional notation, I'd recommend using something like:

```suggestion
          path: spec.template.spec.containers[0].env[?(@.name == "SYSDIG_ACCESS_KEY")]
```

This applies also to the other assertion at line 32

charts/cluster-scanner/tests/secret_test.yaml
```diff
@@ -60,3 +60,45 @@ tests:
...
+          path: data.js_tls_ca_paths
+      - not: true
+        isNull:
+          path: data.js_tls_key
+      - not: true
+        isNull:
+          path: data.js_tls_cert
+      - not: true
+        isNull:
+          path: data.js_tls_ca
```
michele-mangili, line 102: I was thinking about the situation where the customer does enable TLS, but he wants to use a certificate they have. In that case I do not expect the `js_tls_key`, `js_tls_cert` and `js_tls_ca` values to be set. Is this right? If so, I'd recommend to add another unit test to cover also this case.

charts/cluster-scanner/templates/_helpers.tpl
```diff
@@ -279,3 +289,54 @@ Return local registry secrets in the correct format: <namespace_name>/<secret_na
...
+
+{{/*
+Return true if any of TLS CA, private Key and certificate have been provided.
+*/}}
+{{- define "cluster-scanner.nats.tls.customTLS" -}}
+    {{- if or (not (empty .Values.runtimeStatusIntegrator.natsJS.tls.key)) (not (empty .Values.runtimeStatusIntegrator.natsJS.tls.cert)) -}}
+        {{- true -}}
+    {{- else -}}
+        {{- false -}}
+    {{- end -}}
```
mech-pig, line 305:
```suggestion
    {{- or (not (empty .Values.runtimeStatusIntegrator.natsJS.tls.key)) (not (empty .Values.runtimeStatusIntegrator.natsJS.tls.cert)) -}}
```
shouldn't this be equivalent?

charts/cluster-scanner/templates/_helpers.tpl
```diff
@@ -112,6 +112,16 @@ ise_js_consumer_ack_wait: "240s"
 ise_js_consumer_max_deliver: "1"
 ise_js_consumer_deliver_policy_all: "true"
 ise_js_producer_subject: "analysis.sboms"
+{{- if .Values.runtimeStatusIntegrator.natsJS.tls.enabled }}
+ise_js_tls_skip_verify: {{ .Values.imageSbomExtractor.natsJS.tls.skipVerify | default "false" | quote }}
+{{- if not .Values.imageSbomExtractor.natsJS.tls.skipVerify }}
+{{- if eq (include "cluster-scanner.nats.tls.customTLS" .) "true" }}
```
mech-pig, line 118: why are you comparing this with a string? doesn't `(include "cluster-scanner.nats.tls.customTLS" .)` return a bool? Mayber we should also rename the definition as `cluster-scanner.nats.tls.hasCustomCert` or `cluster-scanner.nats.tls.customCertDefined`. wdyt?

charts/cluster-scanner/templates/secret.yaml
```diff
@@ -14,6 +14,14 @@ data:
...
   cache_redis_password: {{ .Values.imageSbomExtractor.cache.redis.password | b64enc | quote }}
 {{- end }}
+{{- if .Values.runtimeStatusIntegrator.natsJS.tls.enabled }}
+  {{- if eq (include "cluster-scanner.nats.tls.customTLS" .) "true" }}
+  js_tls_key_path: {{ required "please provide a TLS private key path" .Values.runtimeStatusIntegrator.natsJS.tls.key |  quote }}
+  js_tls_cert_path: {{ required "please provide a TLS certificate path" .Values.runtimeStatusIntegrator.natsJS.tls.cert | quote }}
+  {{- else }}
+  {{- include "cluster-scanner.nats.tls.certificate" . }}
+  {{- end }}
+{{- end }}
```
mech-pig, line 24: Is it ok to treat paths as secrets? I don't know what the best practices are in this case, but maybe we should include here only the real secrets and move the non-secret config to a configmap. I repeat, I'm not an expert, just sharing my thoughts. cc @michele-mangili

mech-pig (commented):
left some comments, but overall lgtm

charts/cluster-scanner/templates/_helpers.tpl
michele-mangili: I agree @mech-pig

charts/cluster-scanner/templates/deployment.yaml
```diff
@@ -48,6 +48,22 @@ spec:
...
+            secretName: {{ include "cluster-scanner.fullname" . }}
+            items:
+              - key: js_tls_key
+                path: key.pem
+              - key: js_tls_cert
+                path: cert.pem
+              {{- if not .Values.imageSbomExtractor.natsJS.tls.skipVerify }}
+              - key: js_tls_ca
+                path: ca.pem
+              {{ end }}
```
michele-mangili, line 64:
```suggestion
              {{- end -}}
```

charts/cluster-scanner/templates/secret.yaml
michele-mangili: I agree @mech-pig

charts/cluster-scanner/templates/_helpers.tpl
michele-mangili: FYI: https://blog.palark.com/advanced-helm-templating/ (1.6)

> When used in conditions of if blocks, include returns strings only and does not convert them to other data types. In other words, if the boolean true is returned from the template, it becomes the string "true". Any non-empty string is equivalent to boolean true in the if block condition. Note that if the boolean false is returned from the template, it becomes a non-empty "false" string, which itself means a boolean true.
>
> If you want to get a real boolean false in the if condition, you can return an empty string from the template:

```
{{- define "returnPseudoBoolean" }}
  {{- if eq . "pleaseReturnTrue" }}
true
  {{- else if eq . "pleaseReturnFalse" }}
  {{- end }}
{{- end }}
```

charts/cluster-scanner/templates/_helpers.tpl
michele-mangili: Unfortunately, it is not. Each definition is treated as a string, and the `"false"` string would evaluate to `true`, see my comment https://github.com/sysdiglabs/charts/pull/1366#discussion_r1345803757.

I'll just simplify this condition, and make it return an empty string if it evaluates to `false`.

PR #1372, opened 2023-09-26
feat(cluster-scanner): added configuration for docker registry mirrors

michele-mangili:

As per the title

## Checklist


charts/cluster-scanner/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: Sysdig Cluster Scanner
 
 type: application
 
-version: 0.6.0
+version: 0.6.1
```
mavimo: Since is a new feature I suggest to bump the minor release instead of the patch release here.

charts/cluster-scanner/README.md
```diff
@@ -134,6 +134,7 @@ The following table lists the configurable parameters of the `cluster-scanner` c
...
+| imageSbomExtractor.registryMirrorsConfig           | Provide optional mirrors configuration to be used by Image SBOM Extractor to pull images. [Only Docker HUB images](https://docs.docker.com/registry/recipes/mirror/#gotcha) are going to be pulled from the provided mirrors. When provided, the parameter must be a string in docker-daemon configuration file format (see: https://docs.docker.com/registry/recipes/mirror/#configure-the-docker-daemon). <br> Example: <br> `registryMirrorsConfig: \|-` <br> `{` <br> `"registry-mirrors": [` <br> `"insecure.mirror.acme.com",` <br> `"secure.mirror.acme.com"` <br> `],` <br> `"insecure-registries": [` <br> `"insecure.mirror.acme.com"` <br> `]` <br> `}`                                                      | <code></code>                                 |
```
mavimo: What do you think to set is as yaml object instead of json string and then encode the output at the configmap level? This will make way easier for customer to do manipulation plus will allow us to introduce validation rules using a `values.schema.json` file.

charts/cluster-scanner/Chart.yaml
michele-mangili: Sure, will do, thanks!

charts/cluster-scanner/README.md
michele-mangili: I agree that would make things a lot easier for the customer. I am going to update the chart accordingly. Thank you @mavimo !

mech-pig (approved):

charts/cluster-scanner/values.yaml
```diff
@@ -218,6 +218,25 @@ imageSbomExtractor:
     #   sentinelAddress: ""
     #   ttl: "168h" # Default ttl is 7 days
 
+  # Provide optional registry mirrors configuration to be used by Image SBOM Extractor to pull images.
+  # [Only Docker HUB images](https://docs.docker.com/registry/recipes/mirror/#gotcha) are going to be pulled from
+  # the provided mirrors.
```
aroberts87, line 223: The wording here is a bit confusing for me, but perhaps it's just a lack of expertise in this area.
> Provide optional registry mirrors configuration ... to pull images. Only Docker HUB images are going to be pulled ...

☝🏻 Does this mean any provided registries that are not docker hub will be ignored?

aroberts87 (approved):
lgtm, @michele-mangili! just one minor nit that may just be a result of my lack of knowledge in this area. thanks for addressing the earlier concerns by @mavimo as well. 🙂

charts/cluster-scanner/values.yaml
michele-mangili: Thanks @aroberts87 for your comment!

>☝🏻 Does this mean any provided registries that are not docker hub will be ignored?

Not exactly, let me make an example:

Assume you want to pull `alpine:latest` and you have configured `mirror.mycompany.com` as a mirror. Given that this is an image served by Docker Hub, we will contact the mirror you configured and get the image from there. The mirror will only pull the image from Docker Hub if it does not have a local copy yet.

On the other hand, assume you want to pull `quay.io/sysdig/admission-controller:master`. This comes from `quay.io` not Docker Hub. In this case, we will not contact the mirror at all, but will directly pull the image from `quay.io`.

This behavior is consistent with what Docker does (see https://docs.docker.com/registry/recipes/mirror/#gotcha):

> It's currently not possible to mirror another private registry. Only the central Hub can be mirrored.

If you have any suggestions for better wording, please let me know!

aroberts87: Ah, perfect! Thanks for the clarification. 🙂

PR #1373, opened 2023-09-27
fix(agent): add mount for host's `/run` vol for all deployment types

aroberts87:
When installing on GKE Autopilot clusters with the agent slim mode selected the host's `/run` volume was not being mounted into the agent container. This was preventing the correct retrieval of container metrics.

Since the problematic scenario was the only case where the host's `/run` volume was not being mounted into the agent container, that volume has now been elevated to the "always requested" section.

## Checklist


AlbertoBarba (approved):

mavimo (approved):
We manually tested and now work as expected.

PR #1379, opened 2023-09-28
fix(admission-controller): not detecting events in its own namespace

tembleking:
If we block audit events in the namespace, the AC is not able to detect events in its own namespace.

### DIFF
--- a/charts/admission-controller/Chart.yaml
+++ b/charts/admission-controller/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: admission-controller
 description: Sysdig Admission Controller using Sysdig Secure inline image scanner
 type: application
-version: 0.14.8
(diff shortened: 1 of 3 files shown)


airadier:
This reverts the changes on https://github.com/sysdiglabs/charts/pull/1316, reported in https://github.com/sysdiglabs/charts/issues/1310

So we fix one case and break another one. 

Can we have the namespace selector with a chart flag so it can be set or not, so the customer in #1310 can enable the namespace selector if needed?

tembleking:
Since there are two endpoints, one to validate deployments and another one for k8s audit, for the customer in #1310 would be maybe easier to install the AC with Audit disabled, and then update it to enable the audit, so the ValidatingWebhookConfiguration is updated.

miketnt (approved):

PR #1388, opened 2023-10-02
docs(registry-scanner): specify the namespace to use during installation

mavimo:
## Checklist


AlbertoBarba (approved):

airadier:
@mavimo @AlbertoBarba , hopefully the tests should pass now after merging master. We disabled 2 tests for JFrog Artifactory SaaS because the test instance has been destroyed and they were failing.

AlbertoBarba (commented):
Even if u only modify the README you currently need to bump the chart version.

cc @aroberts87 @mavimo

airadier:
> Even if u only modify the README you currently need to bump the chart version.
> 
> cc @aroberts87 @mavimo

That's right, now a version bump is required because of the latest changes happening in master, but at least the tests should be passing.

airadier (approved):

PR #1406, opened 2023-10-17
feat(node-analyzer, runtime-scanner): runtime scanner ask always for profiling data [SSPROD-31183]

jprieto92:
## Checklist


giuse-sysdig (changes requested):
What happens if **eve_enabled** goes to always **true** to the customers who directly set to **false**? The requirement is not to change this behavior.
Also for the **eve_integration_enabled** I would keep it **false**, that's linked to the eve_connector for Snyk.

PR #1413, opened 2023-10-18
feat(cluster-scanner): Enable platform services only in regions which support them

FedericoFeresini:


charts/cluster-scanner/templates/_helpers.tpl
```diff
@@ -207,10 +207,14 @@ Define the proper imageRegistry to use for imageSbomExtractor
 
 {{/*
 Generates configmap data to enable platform services if onPremCompatibility version is not set, or it is greater than 6.6.0
+It also makes sure that the plarform services are enabled in regions which support them.
 */}}
 {{- define "cluster-scanner.enablePlatformServicesConfig" -}}
 {{- if ( semverCompare ">= 6.6.0" (.Values.onPremCompatibilityVersion | default "6.6.0" )) -}}
+    {{- $regionsPlatformEnabled := dict "us1" "" "us2" "" "us3" "" "au1" "" "eu1" -}}
+    {{- if hasKey  $regionsPlatformEnabled .Values.global.sysdig.region -}}
```
michele-mangili: I would recommend you to run this check only if `onPremCompatibilityVersion` has not been set.

michele-mangili (approved):

PR #1458, opened 2023-11-06
feat(agent): Add Local Forwarder support for Agent

aroberts87:
* Create a new configmap for the local forwarder
* Add local forwarder elements to json schema
* Add volume mapping for new configmap

## Checklist


charts/agent/templates/_helpers.tpl
```diff
@@ -506,3 +506,8 @@ sysdig_capture_enabled: false
 true
 {{- end }}
 {{- end }}
+
+{{/* Return the name of the local forwarder configmap */}}
+{{- define "agent.localForwarderConfigMapName" }}
+{{- include "agent.configmapName" . | printf "%s-local-forwarder" | trunc 63 | trimSuffix "-" }}
```
mavimo: If the agent configmapName is loger we will cut out the `local-forwarder` suffix creating clash, I suggest to cut the previous part and then append the suffix, like:
```suggestion
{{- include "agent.configmapName" . | trunc 46 | trimSuffix "-" | printf "%s-local-forwarder" }}
```

charts/agent/values.schema.json
```diff
@@ -23,6 +24,100 @@
...
       }
+    },
+    "localForwarder": {
+      "type": "object",
+      "required": [
+        "enabled",
+        "integrations",
+        "transmitMessageTypes"
+      ],
+      "if": {
```
mavimo, line 35: FYI: Not sure if the "if" here is supported on old version of `helm` CLI

charts/agent/values.schema.json
```diff
@@ -23,6 +24,100 @@
...
+                "SECURE_AUDIT"
+              ]
+            }
+          }
+        }
+      }
+    }
+  },
+  "$defs": {
+    "integration": {
```
mavimo: let's use a less ambiguous name here ;) 
```suggestion
    "AgentLocalForwarderIntegration": {
```

mavimo (changes requested):

charts/agent/values.yaml
```diff
@@ -280,6 +280,13 @@ tolerations:
     value: "true"
 leaderelection:
   enable: false
+localForwarder:
+  enabled: false
+  transmitMessageTypes:
+    - AUDIT_TAP
```
aroberts87: This needs to be removed.

claffin:
Can we include these in the readme as well? @aroberts87

aroberts87:
> Can we include these in the readme as well? @aroberts87

Absolutely, @claffin. I'm reviving this PR now and will include the entries in the readme.

charts/agent/values.schema.json
aroberts87: It appears to be...

I've been able to successfully use this change with helm 3.6.3 and 3.10.1. I can look into some other approaches here if you'd like to remove the if/then flow though.

charts/agent/values.schema.json
aroberts87: Handled in 372a28b

mavimo (approved):

PR #1501, opened 2023-11-29
feat(node-analyzer): In-USE (EVE) enabled by default

airadier:

In-Use (EVE) will now be GA and officially enabled by default for all customers.

With this PR we change the `eveEnabled` setting to true by default, and we fix an issue where EVE integrations where only enabled if `eveConnector` component was also deployed. This is not necessary, as some integrations happen via a public API, but we need to make sure that `EVE_INTEGRATION_ENABLED` environment variable is also set to true, or the public API won't have any data.

Setting the PR as draft until dependent components are GA.

## Checklist


charts/node-analyzer/Chart.yaml
```diff
@@ -1,9 +1,8 @@
 apiVersion: v2
 name: node-analyzer
 description: Sysdig Node Analyzer
-
 # currently matching Sysdig's appVersion 1.14.34
-version: 1.19.2
+version: 1.19.3
```
mavimo: Please do at least a minor bump here

FedericoFeresini (approved):

PR #1504, opened 2023-11-30
fix(agent): allows to specify non semver tags [SMAGENT-6093] 

AlbertoBarba:

Due to the introduction of some features that are available since version `12.18.0` we introduced a check on the version which fails if the provided tag is not a valid semantic version.

## Checklist

### DIFF
--- a/charts/agent/templates/daemonset.yaml
+++ b/charts/agent/templates/daemonset.yaml
@@ -219,7 +219,7 @@ spec:
               value: /opt/draios/etc/ca-certs/{{- include "sysdig.custom_ca.keyName"  (dict "global" .Values.global.ssl "component" .Values.ssl) -}}
             {{- end }}
           readinessProbe:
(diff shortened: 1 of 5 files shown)


mavimo (approved):
LGTM

charts/agent/templates/daemonset.yaml
```diff
@@ -219,7 +219,7 @@ spec:
               value: /opt/draios/etc/ca-certs/{{- include "sysdig.custom_ca.keyName"  (dict "global" .Values.global.ssl "component" .Values.ssl) -}}
             {{- end }}
           readinessProbe:
-            {{- if ge (semver "12.18.0" | (semver .Values.image.tag).Compare) 0 }}
+            {{- if eq (include "agent.enableHttpProbes" .) "true" }}
             httpGet:
               path: /healthz
```
fuster92, line 224: A totally unrelated comment, is the `z` at the end on purpose?

AlbertoBarba: Yes, it's a sort of convention

PR #1517, opened 2023-12-05
feat(agent): mount agent http-proxy credentials as secret

francesco-furlan:

Mount agent http-proxy credentials as a secret in both agent `DaemonSet` and `Deployment` except when gke autopilot is enabled.

## Checklist

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -30,4 +30,4 @@ sources:
 - https://app.sysdigcloud.com/#/settings/user
 - https://github.com/draios/sysdig
 type: application
-version: 1.17.4
+version: 1.18.0
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -30,4 +30,4 @@ sources:
 - https://app.sysdigcloud.com/#/settings/user
 - https://github.com/draios/sysdig
 type: application
-version: 1.17.3
+version: 1.17.4
--- a/charts/agent/templates/_helpers.tpl
+++ b/charts/agent/templates/_helpers.tpl
@@ -183,6 +183,16 @@ Return the default only if the value is not defined in sysdig.settings.<agent_se
 The following helper functions are all designed to use global values where
 possible, but accept overrides from the chart values.
 */}}
+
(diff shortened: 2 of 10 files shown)


charts/agent/templates/daemonset.yaml
```diff
@@ -249,6 +249,10 @@ spec:
               name: sysdig-agent-config
             - mountPath: /opt/draios/etc/kubernetes/secrets
               name: sysdig-agent-secrets
+            {{- if (include "agent.httpProxyCredentials" .) }}
```
AlbertoBarba: We should probably perform the same changes in `charts/agent/templates/deployment.yaml`

francesco-furlan: uh thanks, I was missing that! 🙏

charts/agent/Chart.yaml
```diff
@@ -30,4 +30,4 @@ sources:
 - https://app.sysdigcloud.com/#/settings/user
 - https://github.com/draios/sysdig
 type: application
-version: 1.17.3
+version: 1.17.4
```
mavimo: Should be a minor release.

charts/agent/templates/_helpers.tpl
```diff
@@ -183,6 +183,16 @@ Return the default only if the value is not defined in sysdig.settings.<agent_se
 The following helper functions are all designed to use global values where
 possible, but accept overrides from the chart values.
 */}}
+
+{{- define "agent.httpProxyCredentials" -}}
+    {{- if hasKey .Values.sysdig.settings "http_proxy" -}}
+        {{- if and (hasKey .Values.sysdig.settings.http_proxy "proxy_user") (hasKey .Values.sysdig.settings.http_proxy "proxy_password") -}}
+proxy_user: {{ .Values.sysdig.settings.http_proxy.proxy_user | toString | b64enc | quote }}
+proxy_password: {{ .Values.sysdig.settings.http_proxy.proxy_password | toString | b64enc | quote }}
```
mavimo, line 191: I'll not do the encoding here 😅

AlbertoBarba (approved):

PR #1520, opened 2023-12-07
Added documentation for Rancher's containerd path

jasonumiker-sysdig:
When onboarding a new customer that uses Rancher recently we found we needed to cater for the different containerd socket path for the runtime scanner to work.

I am adding a section to the Helm chart's documentation illustrating how we did that to save people the effort of having to work this out themselves in the future. It also illustrates how to do the YAML arrays with the [0] in the --set options which wasn't clear to me at first and may help others.


jasonumiker-sysdig:
Per internal discussion moving this to docs.sysdig.com and closing PR here

PR #1561, opened 2024-01-11
feat(agent): enable falcobaseline for agent version 12.9.x and above

maratsal:
This PR makes sure that falcobaseline, drift_control and drift_killer settings are not disabled if agent version 12.19.0 and above is used.

## Checklist


charts/agent/templates/_helpers.tpl
```diff
@@ -411,6 +411,12 @@ agent config to prevent a backend push from enabling them after installation.
             "secure_audit_streams") }}
             {{- $_ := set $secureConfig $secureFeature (dict "enabled" false) }}
         {{- end }}
+    {{ else if and $secureLightMode (semverCompare ">= 12.19.x" .Values.image.tag) }}
```
AlbertoBarba: @maratsal unfortunately we can't use the "semver*" functions since they'll make everything fail in case a non semantic version is provided. We should probably do something like we did in this PR https://github.com/sysdiglabs/charts/pull/1504

AlbertoBarba (changes requested):

charts/agent/templates/_helpers.tpl
AlbertoBarba: U can add an helper like:
```
{{- define "agent.enableFalcoBaseline" }}
{{- if regexMatch "^v?([0-9]+)(\\.[0-9]+)?(\\.[0-9]+)?(-([0-9A-Za-z\\-]+(\\.[0-9A-Za-z\\-]+)*))?(\\+([0-9A-Za-z\\-]+(\\.[0-9A-Za-z\\-]+)*))?$" .Values.image.tag }}
{{- if semverCompare ">= 12.19.0-0" .Values.image.tag }}
{{- printf "true" -}}
{{- end }}
{{- end }}
{{- end }}
```
And then replace the `(semverCompare ">= 12.19.x" .Values.image.tag)` with `(eq (include "agent.enableFalcoBaseline" .) "true")`

maratsal: I see, let me add that logic. actually we probably can intro new function there that would include that regex in it, what do you think?

maratsal: added additional logic

charts/agent/tests/secure_light_config_test.yaml
```diff
@@ -0,0 +1,237 @@
...
+            memdump:
+              enabled: false
+      - matchRegex:
+          path: data['dragent.yaml']
+          pattern: |-
+            network_topology:
+              enabled: false
+    template: configmap.yaml
+
+  - it: Testing if certain settings set to false for agent version 12.20.1
```
AlbertoBarba: Duplicated (see the previous one)

AlbertoBarba (changes requested):

charts/agent/tests/secure_light_config_test.yaml
maratsal: removed duplicate. thanks.

AlbertoBarba (approved):

PR #1593, opened 2024-02-06
fix(agent, kspm-collector, rapid-response, node-analyzer): Set test rollout namespace based on required service accounts

IgorEulalio:
## Checklist

- [ x ] Title of the PR starts with type and scope, (e.g. feat(agent,node-analyzer,sysdig-deploy):)
- [ x ] Chart Version bumped for the respective charts
- [ x ] Variables are documented in the README.md (or README.tpl in some charts)
- [ x ] Check GithubAction checks (like lint) to avoid merge-check stoppers
- [ x ] All test files are added in the tests folder of their respective chart and have a "_test" suffix


charts/admission-controller/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: admission-controller
 description: Sysdig Admission Controller using Sysdig Secure inline image scanner
 type: application
-version: 0.14.21
+version: 0.14.22
```
AlbertoBarba: This bump shouldn't be needed

charts/agent/templates/tests/test-rollout.yaml
```diff
@@ -3,6 +3,7 @@ kind: Pod
 metadata:
```
AlbertoBarba, line 3: You need to bump the agent chart version

AlbertoBarba (changes requested):
Hey @IgorEulalio, i suppose the aim for this PR is to set clearly the namespace where those resources will be defined. If that's so you should have put the `namespace` key inside the `metadata` object while, right now, u did put it in `labels` 😅

AlbertoBarba (approved):

PR #1596, opened 2024-02-07
feat(cluster-scanner,sysdig-deploy): add logic to make it possible to disable platform services if needed.

maratsal:
## Checklist


charts/cluster-scanner/values.yaml
```diff
@@ -322,3 +322,6 @@ ssl:
     existingCaConfigMap:
     # Provide the filename that is defined inside the existing ConfigMap
     existingCaConfigMapKeyName:
+
+# Option to make it possible to disable platform services
+enablePlatformServices: true
```
airadier: we have `disablePlatformScanning` 
 in registry-scanner chart. Should we keep this consistent? https://github.com/sysdiglabs/charts/blob/9e300673a0b32de905fe3d0c8f4ffdef41f00890/charts/registry-scanner/values.yaml#L109
 
 So I would prefer the one in registry-scanner 🙏

airadier (changes requested):

charts/cluster-scanner/values.yaml
maratsal: Thank you! updated code.

airadier (approved):

PR #1651, opened 2024-03-15
feat(admission-controller): Support for VM checks

ProvoK:

This PR enables the support for the VM checks in the AdmissionControllerWebhook


charts/admission-controller/templates/_helpers.tpl
```diff
@@ -395,6 +395,7 @@ webhooks:
     apiVersions: ["v1"]
     operations: ["CREATE", "UPDATE"]
     resources:
+    - "pods"
```
airadier, line 399: do we need to wrap this with the conditional too? Or is it ok to let KSPM just ignore the pods? (They should be eventually supported in the future)

charts/admission-controller/values.yaml
```diff
@@ -195,6 +195,15 @@ webhook:
...
   # The service name for Webhook deployment
   name: webhook
+
+  # Vulnerability Management extension settings
+  vm:
+    enabled: true
+    # NOTE: If you're installing ClusterScanner via ClusterShield
+    # the clusterScannerEndpoint will be in the form `${cluster-shield-installation-fullname}-cluster-scanner:9999`
+    # Port is always 9999
+    clusterScannerEndpoint: "CHANGEME:9999"
```
airadier: Is there a default `cluster-shield-installation-fullname` that we can use instead of CHANGEME? 

I am assuming (CC @guidobonomi ) that installing ClusterScanner via ClusterShield is the only option, because we are not releasing the latest changes for ClusterScanner via standalone chart? If we, I would remove the conditional "If you are installing...".

airadier (approved):

charts/admission-controller/values.yaml
ProvoK: @airadier no cluster shield is not the only option, and (for cluster-shield) unfortunately there's not a sensible default since the service name depends on the (helm) installation name

charts/admission-controller/templates/_helpers.tpl
ProvoK: IIRC (@dimar-apolicy correct me if I'm wrong) KSPM will ignore the pods events as of now.

I can put a guard on that too, though

ProvoK: done @airadier

charts/admission-controller/README.md
```diff
@@ -141,10 +141,12 @@ The following table lists the configurable parameters of the `admission-controll
...
+| webhook.vm.clusterScannerEndpoint                  | NOTE: If you're installing ClusterScanner via ClusterShield the clusterScannerEndpoint will be in the form `${cluster-shield-installation-fullname}-cluster-scanner:9999` Port is always 9999                                                                                                                                                                                                                                                                 | <code>"CHANGEME:9999"</code>                                                                                                                                                                       |
```
airadier: Is this something that needs to be manually configured by the customer? Can we have a safe default instead of CHANGEME:9999?

charts/admission-controller/templates/webhook/clusterrole.yaml
```diff
@@ -18,7 +18,13 @@ rules:
   - pods
 {{- if .Values.webhook.acConfig }}
   - configmaps
+{{- end }}
+{{- if or .Values.webhook.acConfig (.Values.webhook.vm).enabled  }}
```
airadier, line 22: Just curious, why do we use `(.Values.webhook.vm).enabled` in some cases but no parenthesis i.e. in `.Values.webhook.vm.clusterScannerEndpoint` ?

charts/admission-controller/templates/webhook/configmap.yaml
```diff
@@ -42,3 +42,12 @@ data:
   WORKERS: "2"
   WORKER_VERIFYSSL: "true"
   {{- end}}
+  {{- if (.Values.webhook.vm).enabled }}
+  PLUGINS_VM_ENRICHER_IMAGEMETADATA: "{}"
+  PLUGINS_VM_ENRICHER_K8S_LABELS: "{}"
+  PLUGINS_VM_EXECUTOR_EXTRACT_SBOM: |-
+    {
+      "rsiHost": {{ .Values.webhook.vm.clusterScannerEndpoint | default "clusterscanner-cluster-scanner:9999" | quote }},
```
airadier: Ah this seems to be the default if not specified. Is this the default name when deploying the cluster-scanner in the same cluster using the default values or documentation / onboarding commands?

charts/admission-controller/values.yaml
```diff
@@ -188,13 +188,22 @@ webhook:
...
```
airadier: Do we need to use `CHANGEME:9999` as the default value? I see there is a default fallback value:
`rsiHost": {{ .Values.webhook.vm.clusterScannerEndpoint | default "clusterscanner-cluster-scanner:9999" | quote }},`

so maybe clusterScannerEndpoint should be empty or commented in the values.yaml?

ProvoK: Actually, it would be better to not default to anything in here because sysdig-deploy and cluster-shield have different default values, and some are dependant on the input `values.yaml`

charts/admission-controller/templates/webhook/configmap.yaml
ProvoK: it's the default for the cluster-scanner chart, would be probably better to remove this default actually

charts/admission-controller/templates/webhook/clusterrole.yaml
ProvoK: parenthesis are for safety access to nested values that might not exist 😉 

so in the `if` conditions they are used, inside the if block it's not useful

charts/admission-controller/README.md
ProvoK: not really, see https://github.com/sysdiglabs/charts/pull/1651#discussion_r1541502169

dimar-apolicy (approved):

charts/admission-controller/README.md
ProvoK: will add a specific check for this then, as agreed by voice

PR #1679, opened 2024-04-03
feat(node-analyzer,sysdig): Update legacy engine NIA/HostAnalyzer components with security updates

FedericoFeresini:


charts/sysdig/Chart.yaml
```diff
@@ -15,4 +15,4 @@ name: sysdig
 sources:
 - https://app.sysdigcloud.com/#/settings/user
 - https://github.com/draios/sysdig
-version: 1.16.25
+version: 1.16.26
```
michele-mangili: I think we shall not update anymore the `sysdig` chart since it has been deprecated

FedericoFeresini: Should we ditch it then?

michele-mangili (approved):

michele-mangili (approved):

PR #1686, opened 2024-04-04
feat(agent): revert ensure os-release file is mounted

AlbertoBarba:
## Checklist

### DIFF
--- a/charts/agent/templates/daemonset.yaml
+++ b/charts/agent/templates/daemonset.yaml
@@ -244,6 +244,9 @@ spec:
           {{- /* Always requested */}}
             - mountPath: /host/dev
               name: dev-vol
+            - mountPath: /host/usr
+              name: usr-vol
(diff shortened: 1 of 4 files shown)


charts/agent/templates/daemonset.yaml
```diff
@@ -244,6 +244,9 @@ spec:
           {{- /* Always requested */}}
             - mountPath: /host/dev
               name: dev-vol
+            - mountPath: /host/usr
```
AlbertoBarba, line 247: Since this volume was added for autopilot in https://github.com/sysdiglabs/charts/pull/1625/files#diff-18888d7372d81c06ea1ec38c2606ae16bf04d0248bdb8ad475cbc5cc22bbc364R523 i do expect to not be an issue to always mount it from now on

aroberts87 (approved):

PR #1691, opened 2024-04-11
feat(cluster-scanner,sysdig-deploy): add option to populate custom labels for cluster scanner

maratsal:
Adding labels option to cluster scanner so that customer can deploy it in their environment with OPA agent that controls deployment based on labels. Please check with @a1ex-var1amov if needed.

## Checklist


charts/cluster-scanner/values.yaml
```diff
@@ -344,3 +344,6 @@ ssl:
 
 # Option to make it possible to disable platform services
 disablePlatformScanning: false
+
+# Set Cluster Scanner pod labels
+labels: {}
```
mavimo, line 352: @maratsal thx for your contribution!

My suggestion is to use `podLabels` instead od `labels` as term, this fit with the "standard" suggested by helm on chart creations:

 *  https://github.com/helm/helm/blob/main/pkg/chartutil/create.go#L133

and used by other charts:

 * https://github.com/search?q=podLabels+language%3AYAML+path%3Avalues.yaml&type=code

maratsal: sure, will update it. used this construction just because couple of other charts in our repository is using the same pattern.

dark-vex:
@mavimo @maratsal I've added the full test for the labels (like the other charts) and applied some changes by defining the `podLabels` on the helper

mavimo (approved):

PR #1706, opened 2024-04-19
feat(agent): allow to configure minReadySeconds on the daemonset

AlbertoBarba:

The daemonset RollingUpdate strategy doesn't seems to wait for pods to become ready before proceeding with the next batch in the sequence. By setting the `minReadySeconds` this can be achieved.

## Checklist

### DIFF
--- a/charts/agent/values.yaml
+++ b/charts/agent/values.yaml
@@ -120,6 +120,7 @@ serviceAccount:
   name: null
 daemonset:
   deploy: true
+  minReadySeconds: 1
   # Perform rolling updates by default in the DaemonSet agent
   # ref: https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/
   updateStrategy:
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -30,4 +30,4 @@ sources:
 - https://app.sysdigcloud.com/#/settings/user
 - https://github.com/draios/sysdig
 type: application
-version: 1.22.7
+version: 1.22.8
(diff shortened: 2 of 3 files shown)


charts/agent/values.yaml
```diff
@@ -120,6 +120,7 @@ serviceAccount:
   name: null
 daemonset:
   deploy: true
+  minReadySeconds: 1
```
AlbertoBarba, line 123: @aroberts87 @irozzo-1A @mavimo Setting 1 will change the default behaviour for everyone.
With this in place every update batch will start only once all the pods in the previous one will reach the `Ready` state. I'm not sure if we want this to become the default behaviour from now on. Thoughts?

irozzo-1A: @AlbertoBarba IMO we should first clarify if this is this is the expected behavior. I don't have the impression from what I see in the docs and the code that when keeping the default (0), it should not be waiting for readiness.
 https://github.com/kubernetes/kubernetes/blob/f39ece24b27062a8d9eb791ffebc252ea1449734/staging/src/k8s.io/api/apps/v1/types.go#L246-L250
https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/#performing-a-rolling-update

aroberts87: I agree with @irozzo-1A here. Waiting 1 second after the pod claims ready state shouldn't gain much over just waiting for the pod to be in ready state. The fact that the one second difference here seems to resolve the issue smells to me like there is a bug somewhere else in the system.

AlbertoBarba:
Superseded by https://github.com/sysdiglabs/charts/pull/1710

PR #1743, opened 2024-05-14
feat(agent): privileged flag for Sysdig Agent 

deepskyblue86:
## Checklist


charts/agent/Chart.yaml
```diff
@@ -30,4 +30,4 @@ sources:
 - https://app.sysdigcloud.com/#/settings/user
 - https://github.com/draios/sysdig
 type: application
-version: 1.23.3
+version: 1.23.4
```
mavimo: Bump to minor

charts/agent/templates/_helpers.tpl
```diff
@@ -537,25 +537,83 @@ true
...
-{{- if semverCompare ">= 12.19.0-0" .Values.image.tag }}
+{{- if and (include "agent.isSemVer" .Values.image.tag) (semverCompare ">= 12.19.0-0" .Values.image.tag) }}
 {{- printf "true" -}}
 {{- end }}
 {{- end }}
+
+{{- define "agent.annotationsSection" }}
+  {{- if (include "agent.gke.autopilot" .) }}
+    {{- printf "annotations:" }}
+    {{- printf "autopilot.gke.io/no-connect: \"true\"" | nindent 2 }}
```
mavimo: We can avoid to use printf here as it make it more readable:
```suggestion
annotations:
  autopilot.gke.io/no-connect: "true"
```

charts/agent/templates/_helpers.tpl
```diff
@@ -537,25 +537,83 @@ true
...
+  {{- else if or (.Values.daemonset.annotations) (eq "false" (include "agent.privileged" .)) }}
+    {{- printf "annotations:" }}
+    {{- if (eq "false" (include "agent.privileged" .)) }}
+      {{- printf "container.apparmor.security.beta.kubernetes.io/sysdig: unconfined" | nindent 2 }}
+    {{- end }}
+    {{- if .Values.daemonset.annotations }}
+      {{- toYaml .Values.daemonset.annotations | nindent 2 }}
+    {{- end }}
+  {{- end }}
```
mavimo: Same as before, avoid to use printf when we can avoid it.

charts/agent/templates/securitycontextconstraint.yaml
```diff
@@ -15,7 +15,8 @@ allowHostPID: true
 allowHostPorts: false
 allowPrivilegeEscalation: true
 allowPrivilegedContainer: true
-allowedCapabilities: []
+allowedCapabilities:
+- '*'
```
mavimo: I suppose we should use `{{- if eq "true" (include "agent.privileged" .) }}` as well and set the proper capabilities when needed instead of adding `*`

charts/agent/values.yaml
```diff
@@ -173,6 +173,8 @@ ebpf:
   enabled: false
   # Define the kind of eBPF driver that will be used by the agent. Can be `legacy_ebpf` or `universal_ebpf`
   kind: legacy_ebpf
+# Run the Sysdig Agent container as privileged. When set to false, eBPF must be enabled
+privileged: true
```
mavimo, line 204: Let's add at least the values.schema.json to report that we have this field and must be a boolean.

charts/agent/templates/_helpers.tpl
```diff
@@ -537,25 +537,83 @@ true
...
+      - image tag not semver and has 'dev' in it: go on at user's risk
+*/}}
+{{- define "agent.privileged" }}
+  {{- if or .Values.privileged (include "agent.gke.autopilot" .) }}
+    {{- /* OK */ -}}
+    {{- print "true" }}
+  {{- else }}
+    {{- $errMsg := "Disabling 'privileged' flag requires eBPF and Sysdig Agent 13.3.0 or newer" }}
+    {{- /* eBPF is mandatory */ -}}
+    {{- if not (eq "true" (include "agent.ebpfEnabled" .)) }}
```
mavimo, line 644: I think we should check for Universal eBPF instead of all the eBPF flavour here...

charts/agent/templates/_helpers.tpl
```diff
@@ -537,25 +537,83 @@ true
...
+        {{- fail $errMsg }}
+      {{- end }}
+    {{- /* Check for dev version */ -}}
+    {{- else if or (eq "dev" .Values.image.tag) (regexMatch "^[0-9]+\\.[0-9]+\\.[0-9]+dev-\\w+$" .Values.image.tag) }}
+      {{- /* Let it go through  */ -}}
+      {{- print "false" }}
+    {{- else }}
+      {{- /* FAIL */ -}}
+      {{- fail $errMsg }}
+    {{- end }}
```
mavimo: If is NOT a semver tag we should just accept the input as per customer definition.

mavimo (changes requested):

charts/agent/Chart.yaml
deepskyblue86: I've seen it being bumped to patch so far for similar changes, is there a guideline I can refer to?

charts/agent/templates/_helpers.tpl
deepskyblue86: IMHO that's debatable, I find it more readable with printf, which also allows me to be consistent with the line where I have to use `toYaml`

charts/agent/templates/_helpers.tpl
deepskyblue86: Why is that so? `legacy_ebpf` can run with specific capabilities

charts/agent/templates/_helpers.tpl
deepskyblue86: I preferred to do that only with `dev`, to avoid potentially not working configuration to go through with unknown tags. Would you mind elaborating more?

charts/agent/values.yaml
deepskyblue86: There are plenty of values not covered in the schema, any specific reason to add this one?

charts/agent/Chart.yaml
mavimo: We usually refer to use:
 * Bug and when there are no changes in functionality (e.g. new doc) => patch
 * New feature => minor

charts/agent/values.yaml
mavimo: We added it a few months ago (with the Universal eBPF) and we are adding new params, we don't go back and fill all the existing one ATM

charts/agent/Chart.yaml
deepskyblue86: I didn't observe such a pattern on this chart but no problem :+1:

charts/agent/templates/_helpers.tpl
deepskyblue86: Accept as user's risk as requested

charts/agent/templates/securitycontextconstraint.yaml
```diff
@@ -15,7 +15,8 @@ allowHostPID: true
 allowHostPorts: false
 allowPrivilegeEscalation: true
 allowPrivilegedContainer: true
-allowedCapabilities: []
+allowedCapabilities:
+{{- include "agent.capabilities" . }}
```
mavimo: Please make it conditional so we include the capabilities only when needed, keeping the empty list in other conditions.

charts/agent/templates/_helpers.tpl
mavimo: FYI is the same approach we use in [all our charts](https://github.com/search?q=repo%3Asysdiglabs%2Fcharts+annotations&type=code&p=5) and is also the [default behaviour from helm](https://github.com/helm/helm/blob/main/pkg/chartutil/create.go#L259-L262) (so is a well known pattern in the ecosystem), honestly I'll like to keep consistency here. 🙏

mavimo (approved):

PR #1791, opened 2024-06-26
feat(admission-controller,sysdig-deploy): add initial support for cert-manager

### LINKED ISSUE #1761: Support using cert-manager for admission controller certs
It would be great if we could use cert-manager to provision the certs for the admission controller. KEDA supports this, for example: https://github.com/kedacore/charts/tree/main/keda/templates/cert-manager

I'll plan to prepare a PR for review.

yurrriq:

Fixes #1761

## Checklist


charts/admission-controller/templates/_helpers.tpl
```diff
@@ -411,7 +411,9 @@ webhooks:
       name: {{ include "admissionController.webhook.fullname" . }}
       path: /validate
       port:  {{ .Values.webhook.v2.service.port }}
-    caBundle: {{ .Cert }}
+    {{- with .Cert }}
```
yurrriq, line 414: This was the laziest way I could think of to avoid reimplementing this template.

charts/admission-controller/templates/webhook/admissionregistration.yaml
```diff
@@ -1,3 +1,19 @@
+{{- if .Values.webhook.ssl.certManager.enabled }}
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ include "admissionController.webhook.fullname" . }}
+    helm.sh/hook: "post-install, post-upgrade"
```
yurrriq, line 8: I'm not sure what the thinking is behind all the `$existingVac` logic, so I just skipped it here.

charts/admission-controller/templates/webhook/admissionregistration.yaml
```diff
@@ -1,3 +1,19 @@
...
+    meta.helm.sh/release-name: {{ .Release.Name }}
+    meta.helm.sh/release-namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/managed-by: Helm
+  name: {{ include "admissionController.webhook.fullname" . }}
+  namespace: {{ include "admissionController.namespace" . }}
+{{- include "admissionController.webhookTemplate" . }}
```
yurrriq, line 15: Note that I don't inject any `Cert` value here so the resulting `ValidatingWebhookConfiguration` manifest won't contain any `caBundle` as per the `_helper.tpl` changes above.

charts/admission-controller/templates/webhook/certificate.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if .Values.webhook.ssl.certManager.enabled }}
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  labels:
+    {{- include "admissionController.webhook.labels" . | nindent 4 }}
+  name: {{ include "admissionController.webhook.fullname" . }}
+  namespace: {{ include "admissionController.namespace" . }}
+spec:
```
yurrriq, line 9: There's a lot we _could_ parameterize here, but I figure we should start simple and can always iterate.

charts/admission-controller/templates/webhook/certificate.yaml
```diff
@@ -0,0 +1,24 @@
...
+  dnsNames:
+  - {{ include "admissionController.webhook.fullname" . }}
+  - {{ include "admissionController.webhook.fullname" . }}.{{ include "admissionController.namespace" . }}.svc
+  issuerRef:
+    {{- with .Values.webhook.ssl.certManager.issuer.group }}
```
yurrriq, line 14: I've deliberately done it this we so we can easily add a boolean parameter like `webhook.ssl.certManager.issuer.create` later to determine whether we should create an issuer.

yurrriq (commented):
Given an existing `ClusterIssuer/selfsigning-isser`:

```yaml
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: selfsigning-issuer
spec:
  selfSigned: {}
```

... and the following values to a release of `sysdig-deploy`:

```yaml
admissionController:
  enabled: true
  webhook:
    ssl:
      certManager:
        enabled: true
        issuer:
          name: selfsigning-issuer
```

... the `sysdig-deploy` and `admission-controller` charts do what I'd expect:

* create a `Certificate` for the `ValidatingWebhookConfiguration`, using the existing `ClusterIssuer/selfsigning-issuer`
* don't include any `caBundle` in the `webhooks`
* annotate the `ValidatingWebhookConfiguration` so `cainjector` will inject the cert bundle

airadier:
CC @mavimo @AlbertoBarba which team should review and merge? I assume we want to port the changes to the cluster-shield chart too

airadier:
@yurrriq thanks so much for this contribution! We are going to review internally and also port the changes to the cluster-shield chart, as this is now the official deployment method for Admission Controller.

Thanks again.

PR #1801, opened 2024-07-08
feat(agent): add /lib64 volume mount

francesco-furlan:

Add optional host `/lib64` to `/host/usr/lib64` mount (except for gke autopilot), since in some environment the proper `ld-linux-*` is under `/lib64`

## Checklist

### DIFF
--- a/charts/agent/templates/daemonset.yaml
+++ b/charts/agent/templates/daemonset.yaml
@@ -298,9 +298,6 @@ spec:
               name: vardata-vol
             - mountPath: /host/var/run
               name: varrun-vol
-            - mountPath: /host/usr/lib64
-              name: lib64-vol
-              readOnly: true
             {{- if (include "agent.ebpfEnabled" .) }}
             - mountPath: /root/.sysdig
               name: bpf-probes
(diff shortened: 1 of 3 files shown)


charts/agent/templates/daemonset.yaml
```diff
@@ -315,6 +321,9 @@ spec:
               name: vardata-vol
             - mountPath: /host/var/run
               name: varrun-vol
+            - mountPath: /host/usr/lib64
+              name: lib64-vol
+              readOnly: true
```
iurly: If I get this right, this is for the `agent-slim` container. Shouldn't the volume mount only be needed for `agent-kmodule` (`initContainer`)? If that's the case, this mount could be removed.

iurly (approved):
In my understanding, there's one mount that should be removed.
Plus, I don't understand how this mount is optional. Could you please elaborate?

charts/agent/templates/daemonset.yaml
francesco-furlan: I need to re-test this but if I recall correctly we need it for ebpf, updates soon

francesco-furlan: after some tests, I removed it from the main container, `initContainer` should be sufficient. Thanks for the comment!

francesco-furlan:
> I don't understand how this mount is optional. Could you please elaborate?

Ouch I thought I added the `optional: true` flag, but I didn't and btw it's only available for secrets volume and not `hostPath`.
Anyway the default behaviour is that if the volume doesn't exists, it creates an empty dir; that for our use-case is ok.

PR #1871, opened 2024-08-02
fix(agent): use dig function for prometheus_exporter settings check

aroberts87:
Migrate the new prometheus_exporter port creation logic to use Helm's '[dig](https://helm.sh/docs/chart_template_guide/function_list/#dig)' function to protect against cases where the values may not be present.

## Checklist

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -30,4 +30,4 @@ sources:
 - https://app.sysdigcloud.com/#/settings/user
 - https://github.com/draios/sysdig
 type: application
-version: 1.27.14
+version: 1.27.15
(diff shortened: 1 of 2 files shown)


francesco-furlan:
@aroberts87 when was `dig` introduced? Are we sure to not break older versions of helm?

aroberts87:
> @aroberts87 when was `dig` introduced? Are we sure to not break older versions of helm?

@francesco-furlan - The `dig` function was added in [v3.5.0](https://github.com/helm/helm/releases/tag/v3.5.0), which is fortunately below our minimum supported version of v3.6.

francesco-furlan (approved):

mavimo (approved):

PR #1922, opened 2024-09-06
fix(node-analyzer): Reorder volume list

gso-dc:
OpenShift stores the volume list in alphabetical order. To avoid unnecessary reconcile runs when using GitOps - align the order in helm chart with OpenShift expected order.

## Checklist

### DIFF
--- a/charts/node-analyzer/Chart.yaml
+++ b/charts/node-analyzer/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: node-analyzer
 description: Sysdig Node Analyzer
 # currently matching Sysdig's appVersion 1.14.34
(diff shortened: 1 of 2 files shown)


mavimo:
@gso-dc can you please bump also the related chart version in:
 * `https://github.com/sysdiglabs/charts/blob/main/charts/node-analyzer/Chart.yaml#L5`
 
if you don't mind the `cluster-shield` will be handled in a different PR as we use a different workflow for the chart upgrade..

mavimo:
@gso-dc replaced by https://github.com/sysdiglabs/charts/pull/1934

BTW _Cluster Shield_ was released a few days ago

Issue #1944, opened 2024-09-13
sysdig-clustershield deployment does not restart when configmap, secrets, or the webhook change.


yoderme:
Imagine the following scenario: the clustershield helm chart is
    re-deployed and the contents of either its configmap, secrets, or
    webhook change... and the clustershield deployment itself does not
    change. The clustershield deployment won't restart because it does not
    know that anything has changed - even though the things it depended on
    change.

    This problem is especially pronounced for clustershield because with
    the default behavior it will auto-generate the certificates used for
    communication between the webhook and the deployment. If the webhook
    certificate changes and the deployment doesn't restart, the webhook
    won't be able to communicate with the pods in the deployment.

yoderme:
https://github.com/sysdiglabs/charts/pull/1945

PR #1945, opened 2024-09-13
feat: Adding checksum annotations to the clustershield deployment

yoderme:
Imagine the following scenario: the clustershield helm chart is re-deployed and the contents of either its configmap, secrets, or webhook change... and the clustershield deployment itself does not change. The clustershield deployment won't restart because it does not know that anything has changed - even though the things it depended on change.

This problem is especially pronounced for clustershield because with the default behavior it will auto-generate the certificates used for communication between the webhook and the deployment. If the webhook certificate changes and the deployment doesn't restart, the webhook won't be able to communicate with the pods in the deployment.

The canonical way to address this is to put checksums of the depended-upon templates into the deployment as annotations. Now when resources the deployment depend upon change (like the auto-genereted certificates) the clustershield pods will restart.

This is the same solution implemented in the nodeanalyzer daemonset.

## Checklist

### DIFF
--- a/charts/cluster-shield/Chart.yaml
+++ b/charts/cluster-shield/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: cluster-shield
 description: Cluster Shield Helm Chart for Kubernetes
 type: application
-version: 1.3.1
-appVersion: "1.3.1"
+version: 1.3.2
+appVersion: "1.3.2"
 maintainers:
   - name: AlbertoBarba
     email: alberto.barba@sysdig.com
--- a/charts/cluster-shield/Chart.yaml
+++ b/charts/cluster-shield/Chart.yaml
@@ -3,7 +3,7 @@ name: cluster-shield
 description: Cluster Shield Helm Chart for Kubernetes
 type: application
 version: 1.3.2
-appVersion: "1.3.2"
+appVersion: "1.3.1"
 maintainers:
   - name: AlbertoBarba
     email: alberto.barba@sysdig.com
--- a/charts/cluster-shield/templates/deployment.yaml
+++ b/charts/cluster-shield/templates/deployment.yaml
@@ -15,10 +15,13 @@ spec:
       {{- include "cluster-shield.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
       annotations:
-        {{- toYaml . | nindent 8 }}
-      {{- end }}
+        checksum/configmap: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum | quote  }}
+        checksum/secrets: {{ include (print $.Template.BasePath "/secrets.yaml") . | sha256sum | quote  }}
+        checksum/webhook: {{ include (print $.Template.BasePath "/validatingwebhookconfiguration.yaml") . | sha256sum | quote  }}
+        {{- if .Values.podAnnotations }}
+        {{- toYaml .Values.podAnnotations | nindent 8 }}
+        {{- end }}
       labels:
         {{- include "cluster-shield.labels" . | nindent 8 }}
   {{- with .Values.podLabels }}
--- a/charts/cluster-shield/tests/custom_ca_test.yaml
+++ b/charts/cluster-shield/tests/custom_ca_test.yaml
@@ -2,6 +2,8 @@ suite: Test Custom CA
 templates:
   - templates/deployment.yaml
   - templates/secrets.yaml
+  - templates/configmap.yaml
+  - templates/validatingwebhookconfiguration.yaml
 values:
   - ../ci/base-values.yaml
 release:
--- a/charts/cluster-shield/tests/deployment_test.yaml
+++ b/charts/cluster-shield/tests/deployment_test.yaml
@@ -1,6 +1,9 @@
 suite: Test Deployment
 templates:
   - templates/deployment.yaml
+  - templates/secrets.yaml
+  - templates/configmap.yaml
+  - templates/validatingwebhookconfiguration.yaml
 values:
   - ../ci/base-values.yaml
 release:
@@ -33,6 +36,7 @@ tests:
       - contains:
           path: spec.template.spec.containers[?(@.name == "cluster-shield")].args
           content: run-all-namespaced
+    template: templates/deployment.yaml
 
   - it: Test run-all mode
     set:
@@ -57,6 +61,7 @@ tests:
       - contains:
           path: spec.template.spec.containers[?(@.name == "cluster-shield")].args
           content: run-all
+    template: templates/deployment.yaml
 
   - it: Test GKE Autopilot
     set:
@@ -80,6 +85,7 @@ tests:
       - contains:
           path: spec.template.spec.containers[?(@.name == "cluster-shield")].args
           content: run-all
+    template: templates/deployment.yaml
 
   - it: Test accessKeySecret is mounted when provided
     set:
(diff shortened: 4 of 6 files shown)


yoderme:
https://github.com/sysdiglabs/charts/issues/1944

charts/cluster-shield/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: cluster-shield
 description: Cluster Shield Helm Chart for Kubernetes
 type: application
-version: 1.3.1
-appVersion: "1.3.1"
+version: 1.3.2
+appVersion: "1.3.2"
```
mavimo: We are not releasing a new version fo the _Cluster Shield_ but only in the chart so this should not be changed.
```suggestion
appVersion: "1.3.1"
```

mavimo:
@yoderme 
> Imagine the following scenario: the clustershield helm chart is re-deployed and the contents of either its configmap, secrets, or webhook change... and the clustershield deployment itself does not change. The clustershield deployment won't restart because it does not know that anything has changed - even though the things it depended on change.

The Cluster Shield should notice changes in the configmap and secrets and it should pick up the changes in the configuration. It may take some time (as kubernetes will update changes in the pod with some delay) but hash should not be required for that.

Do you have a situation where this is not happening?

yoderme:
> > Imagine the following scenario: the clustershield helm chart is re-deployed and the contents of either its configmap, secrets, or webhook change... and the clustershield deployment itself does not change. The clustershield deployment won't restart because it does not know that anything has changed - even though the things it depended on change.
> 
> The Cluster Shield should notice changes in the configmap and secrets and it should pick up the changes in the configuration. It may take some time (as kubernetes will update changes in the pod with some delay) but hash should not be required for that.
> 
> Do you have a situation where this is not happening?

Thanks for the reply @mavimo !

I read this and said to myself "I've never seen this not happen" and scratched my head for awhile.  I am not sure if the clustershield pod will notice the configmap change (I didn't test it), but it definitely won't notice it when the TLS certificates change.

So I went off to prove this to myself.  I did a helm install
```
helm upgrade --install sysdig sysdig/sysdig-deploy -n sysdig --reuse-values --version 1.65.2
```

and then ran the same command again.  Nothing restarted... but I also noticed that the certificates _also did not change_.  The `sysdig-clustershield-tls-certs` secrets didn't change, and the validating webhook's caBundle didn't change.  This puzzled me until I realized that the helm function `cluster-shield.tlsGenCerts` does a `lookup()` of the certificate and uses it if it exists.  This is clever and avoids the problem.

HOWEVER, if one were to delete `sysdig-clustershield-tls-certs` and then run the `helm upgrade` above again, the TLS certs would change, the validating webhook would change, and the pods would not restart and would not pick up the change.  This leads to lots of the following message in the kube api server logs

```
ailed calling webhook "audit.secure.sysdig.com": failed to call webhook: Post "https://sysdig-clustershield.sysdig.svc:6443/k8s-audit?timeout=5s": tls: failed to verify certificate: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "sysdig-clustershield")
```

The way to get the pods to notice this change is the annotations with the hash.

This is actually a problem for me.  I'm using kustomize basically like so:
```
helmCharts:
  - name: sysdig-deploy
    repo: https://charts.sysdig.com
    version: 1.62.1
    releaseName: sysdig
    namespace: sysdig
    valuesFile: values.yaml
```

On the back end, what happens here is that kustomize calls `helm template` to create a bunch of yaml and then applies that yaml.  Because it runs `helm template` and there is no cluster info given, the TLS cert will change every time, and ... you see where this is headed.

mavimo:
@yoderme thanks for the clarification.

I understand your use case, I'll check but secrets and configmap hashes shouldn't be required (the _Cluster Shiedl_ is running a watch process on the files, and restarting the pod isn't really required).

As for the webhook certificate, I'll check what are the implications are and get back to you.
PS: If you want, you can use an external secret for the cert so that the helm does not create a new cert on every release.

yoderme:
> I understand your use case, I'll check but secrets and configmap hashes shouldn't be required (the _Cluster Shiedl_ is running a watch process on the files, and restarting the pod isn't really required).

Admittedly I have not checked what happens when these change, so OK.  It's the webhook cert that's been the problem for me.

> As for the webhook certificate, I'll check what are the implications are and get back to you. PS: If you want, you can use an external secret for the cert so that the helm does not create a new cert on every release.

Understood about using an external secret.  Plausible, but also a bit of a pain.  Thanks @mavimo for looking at this.  If it's just the web hook cert that's a problem I could change the PR to just have the clustershield pods restart on changes to the validating webhook.  Otherwise you'll have to monitor/restart internally in the clustershield pod.

mavimo:
@yoderme we are doing some internal checks as this may have some implications forcing cert changes which may cause problems in case of re-deployments.
I want to assure you that we are looking into this, but it may take some extra time, sorry for that! We will come back to this PR once we have checked some edge cases and based on our research we may ask you to make some changes to the PR, in the meantime thanks for your patience!

yoderme:
@mavimo understood and thanks for working on this!

yoderme:
I was thinking about this, and realized that the _really cool_ way to deal with this entire problem is to not set up the webhook certificate or the validatingwebhook in helm at all. Instead, when the cluster shield process starts up, have it create a self signed certificate and then talk to the k8s server to create the validating webhook using that cert.  Then on process shutdown remove the validating webhook.  The cluster shield service account will have to have the rbac k8s ability to do this, but that's not hard.  It's a big change from where you are now though. Food for thought.

mavimo:
@yoderme thanks to investing your time thinking about this. I see some drawbacks to what you are suggesting, or at least to making it the default approach:
  * user who rely on external certificate generation (e.g. because they have cert manager or other tools) and then link the chart may not be able to get it generated outiside with their on CA
  * we need to do some extra operation to validate that -in case of autoscaling or multiple replicas- this is only doneby one instance.
  * we may have a scale-down operation so one instance will be terminated, but should not remove the cert/webhook as other instances are still running
  * in case of failure we are not be able to do the proper cleanup if we have  the cluster with "unmanaged" resources
  * for people doing gitops, they will have resources in the cluster that are not defined in the code, causing drift

OFC all these issues can be solved, but we are also investigating to have the hot-reload for certificate as we do for secrets and configmap, but we are not sure if this can be done in a way that does not generate drawbacks...

Some people in the team are joining the investigation, we will post some update here soon

yoderme:
> I see some drawbacks

It's not perfect, but it does have advantages. To your point "user who rely on external certificate generation" - you don't have to tell anyone this certificate exists. ;-). It's only for the purposes of communication between the api server and the pod, so it's really hidden.

For an example of an open source project that takes this approach, see Kyverno.  https://github.com/kyverno/kyverno

AlbertoBarba:
Hi @yoderme, i did perform some tests on your PR and everything seems to be fine.
The cluster-shield application is able to recognise when the configmap or the secrets have been changed, so there should be no need to have the `checksum/configmap` and `checksum/secrets` in place.
I'll only keep the `checksum/webhook` to allow the reload of the certificates.

mavimo:
@yoderme I have a branch where I rebased and make the changes to support webhook, if you can grant me access to the PR  I'll push my update on your branch

yoderme:
> @yoderme I have a branch where I rebased and make the changes to support webhook, if you can grant me access to the PR I'll push my update on your branch

I added you to my repo.  You may just have to take my changes and submit them yourself; that's fine with me, no worries.

mavimo:
> You may just have to take my changes and submit them yourself; that's fine with me, no worries

@yoderme I'll do (I'll try to keep commit attribution to you)

francesco-furlan:
Hi @yoderme, we released Cluster Shield `1.6.0` with the `checksum/webhook` annotation, here the [PR](https://github.com/sysdiglabs/charts/pull/2073/files). Going to close this since should be solved now.
Thanks again for your contribution! 🚀

PR #1968, opened 2024-10-07
fix(shield): add default host shield tolerations and affinity

aroberts87:
The Shield chart was not defining a default set of node affinity and toleration settings which resulted in the Host Shield not being installed on expected nodes.

## Checklist


charts/shield/templates/host/_helpers.tpl
```diff
@@ -54,24 +54,44 @@ If release name contains chart name it will be used as a full name.
...
 {{- end -}}
 
 {{- define "host.affinity" -}}
 {{- $affinity := merge (dict) .Values.affinity .Values.host.affinity -}}
+{{- if $affinity }}
 {{- with $affinity -}}
 {{- . | toYaml -}}
 {{- end -}}
+{{- else -}}
+nodeAffinity:
```
AlbertoBarba: Why don't we set this as default in the values for `host.affinity`?

charts/shield/values.yaml
```diff
@@ -206,7 +206,23 @@ host:
   # The node selector for the host shield
   node_selector: {}
   # The tolerations for the host shield
-  tolerations: []
+  tolerations:
```
AlbertoBarba:
```suggestion
  tolerations: # +doc-gen:break
```
And re-run `make docs`

AlbertoBarba (changes requested):

charts/shield/templates/host/_helpers.tpl
aroberts87: There wasn't a strong reason not to. I was just exploring an option here to not clutter up the default `values.yaml` with verbose default settings. I'm open to moving this out to the file though. What are your thoughts?

AlbertoBarba: The only good point i have in my mind for having it in the values it that the generated documentation will report it as default value, while doing it in this way will hide it in the helpers

aroberts87: That's a good point. 🤔 👍🏻

mavimo: I agree on hiding from the doc (but maybe we may have some annotation for that?) but adding in the default values looks like more clear to users, plus will make the code simpler here  (no if else, no special scenario, ..) 🤔

AlbertoBarba (approved):

PR #1974, opened 2024-10-09
feat(shield): allow custom env vars and volume mounts

aroberts87:
Allow the users to specify custom environment variables and volumes/volume mounts into the Cluster Shield and Host Shield pods/containers.

## Checklist


aroberts87:
@AlbertoBarba - https://github.com/sysdiglabs/charts/commit/7f0a8788368a176381fc1d2d2bdf584415596de6 was added purely to bypass the `helm upgrade` issue caused by the use of `--reuse-values` flag in the CI flows. Realistically this is not a situation we'll have to guard against at the moment due to the chart still being under heavy development. We can remove these in a follow up commit to reduce the cruft in the templates.

AlbertoBarba (approved):

PR #2029, opened 2024-11-13
fix(shield): update custom CA path in host shield config

aroberts87:
The shield chart places custom CA certificates in `/opt/draios/certificates` with a filename that matches either the `ssl.ca.key_name` or `ssl.ca.existing_ca_secret_key_name` values. This change causes the chart to set the host shield's `http_proxy.ca_certificate` setting to match the expected location.

## Checklist

### DIFF
--- a/charts/shield/templates/host/_helpers.tpl
+++ b/charts/shield/templates/host/_helpers.tpl
@@ -24,6 +24,9 @@ If release name contains chart name it will be used as a full name.
     {{- $host := split ":" $parsedProxyConfig.host -}}
     {{- $_ := set $proxyConfig "proxy_host" $host._0 -}}
     {{- $_ := set $proxyConfig "proxy_port" $host._1 -}}
+    {{- if or .Values.ssl.ca.key_name .Values.ssl.ca.existing_ca_secret_key_name }}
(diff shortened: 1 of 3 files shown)


charts/shield/templates/host/_helpers.tpl
```diff
@@ -24,6 +24,9 @@ If release name contains chart name it will be used as a full name.
     {{- $host := split ":" $parsedProxyConfig.host -}}
     {{- $_ := set $proxyConfig "proxy_host" $host._0 -}}
     {{- $_ := set $proxyConfig "proxy_port" $host._1 -}}
+    {{- if or .Values.ssl.ca.key_name .Values.ssl.ca.existing_ca_secret_key_name }}
+      {{- $_ = set $proxyConfig "ca_certificate" (printf "certificates/%s" (include "common.custom_ca.key_name" .)) }}
+    {{- end }}
```
AlbertoBarba:
```suggestion
    {{- if (include "common.custom_ca.enabled" .) }}
      {{- $_ = set $proxyConfig "ca_certificate" (include "common.custom_ca.path" (mergeOverwrite . (dict "CACertsPath" "certificates/"))) }}
    {{- end }}
```

AlbertoBarba (approved):

PR #2038, opened 2024-11-15
feat(shield): enable kubernetes-metadata by default and decrease delegated nodes number to 0

AlbertoBarba:

Enable kubernetes-metadata by default and decrease delegated nodes number to 0

## Checklist


charts/shield/templates/host/_configmap_helpers.tpl
```diff
@@ -100,7 +100,9 @@ true
 {{- $config := dict
   "k8s_cluster_name" .Values.cluster_config.name
   "collector" (include "common.collector_endpoint" .)
-  "collector_port" .Values.sysdig_endpoint.collector.port }}
+  "collector_port" .Values.sysdig_endpoint.collector.port
+  "k8s_delegated_nodes" (dig "k8s_delegated_nodes" 0 .Values.host.additional_settings)
```
aroberts87: I imagine we'd want to have a check to see if `features.kubernetes_metadata.enabled` has been set to `false` here. If this feature is turned off I think we'd like the agent/host shield to run with the default values (unless manually overridden of course).

AlbertoBarba: @mavimo @irozzo-1A WDYT?

irozzo-1A: Yep, I agree that would be better to use the default if `kubernetes_metadata` is disabled

AlbertoBarba: Ok, i did apply the changes. User specified configuration will alway rule but, as a default, if the feature is enabled we do force the delegated agents to 0 and, if not, we simply don't set that configuration.

aroberts87 (approved):

PR #2043, opened 2024-11-19
chore(shield): improve values and their descriptions

AlbertoBarba:
## Checklist


charts/shield/values.yaml
```diff
@@ -152,6 +152,10 @@ features:
 
 host:
   # The driver to use for the host agent (Accepted Values: kmod, legacy_ebpf, universal_ebpf)
+  # Use kmod for kernel version 2.6 to 5.7 where 3rd party modules are allowed.
+  # Use legacy_ebpf for kernel version 4.14 to 5.7 where 3rd party modules are restricted.
+  # Use universal_ebpf for kernel version 5.8 or greater.
+  # KAA Why wouldn't we set the default to universal_ebpf?
```
AlbertoBarba: Remove

aroberts87 (approved):
It is worth noting we're making a breaking change here with the key naming, but we're not yet tech preview with charts so I think we're safe. It would be nice to communicate this though.

PR #2051, opened 2024-11-21
feat(agent): add option to override security.openshift.io/v1 API presence

jsecchiero:
## Checklist


charts/agent/values.yaml
```diff
@@ -404,3 +404,6 @@ tests:
     tag: 1.31.2
 # Allow to modify DNS policy
 dnsPolicy: null
+# Overrides `security.openshift.io/v1` API detection
+# useful while using "helm template" and to generate security context constraints
+hasAPISecurityOpenshiftV1Override: false
```
markmcmurray-sysdig, line 409: Given (as discussed): 
• If .Values.hasAPISecurityOpenshiftV1Override is provided the SCC template will use that value, overriding `(.Capabilities.APIVersions.Has "security.openshift.io/v1")`
• If .Values.hasAPISecurityOpenshiftV1Override is not provided it will use `(.Capabilities.APIVersions.Has "security.openshift.io/v1")`

I would be inclined not to provide a default value here. This means the default behaviour will be to use `(.Capabilities.APIVersions.Has "security.openshift.io/v1")` and only if a value is purposefully provided will we override it.

I don't know if that has documentation implications though (ie will the chart docs not mention this value and essentially make it a secret override only found by reading `charts/agent/templates/securitycontextconstraint.yaml`? Maybe that is a desired side-effect?

PR #2062, opened 2024-12-02
feat(agent): [SMAGENT-8424] Allow for custom security context

iurly:
Introduce a new top-level entry `customSecurityContext` in Values, so to be able to override the securityContext automatically populated by the chart. This is to be used sparingly, and particularly in circumstances where a very harcoded securityContext should be used instead (like in the case of BottleRocket).

## Checklist


charts/agent/README.md
```diff
@@ -173,6 +173,7 @@ The following table lists the configurable parameters of the Sysdig chart and th
...
 | `extraVolumes.mounts`                                   | Specifies the mount points for additional volumes                                                                                                                                        | `[]`                                                                                                      |
 | `extraSecrets`                                          | Allows passing extra secrets that can be mounted via extraVolumes                                                                                                                        | `[]`                                                                                                      |
+| `customSecurityContext`                                 | Allows overriding the securityContext so to overcome limitations and bugs of certain OSes. **WARNING** Only set this as the last resort and only for very well-known issues              | `[]`                                                                                                      |
```
aroberts87: I would personally try to make this sound a little less scary (even though you're right to discourage its use). What about this?

```
Allows overriding the Security Context of the Agent pod(s). Primarily intended resolving targeted incidents and debugging.
```

charts/agent/tests/security_context_test.yaml
```diff
@@ -0,0 +1,140 @@
...
+      privileged: true
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[?(@.name == "sysdig")].securityContext
+          value:
+            allowPrivilegeEscalation: true
+            privileged: true
+            readOnlyRootFilesystem: false
+            runAsNonRoot: false
+            runAsUser: 0
```
aroberts87: This seems to be a duplicate of the test before it.

charts/agent/tests/security_context_test.yaml
```diff
@@ -0,0 +1,140 @@
...
+      - isNotSubset:
+          path: spec.template.spec.containers[?(@.name == "sysdig")].securityContext
+          content:
+            allowPrivilegeEscalation: false
+            privileged: true
+            readOnlyRootFilesystem: false
+            runAsNonRoot: false
+            runAsUser: 0
+
+  - it: Ensure the customSecurityContext (mainly for bottlerocket) is honored
```
aroberts87: jm2c, we may want to remove the reference to bottlerocket here, no matter how true it may be.

charts/agent/README.md
iurly: ok

charts/agent/tests/security_context_test.yaml
iurly: good catch, thanks! removed

charts/agent/tests/security_context_test.yaml
iurly: uhm, ok, you probably have good reasons to say so.
done.

charts/agent/templates/_helpers.tpl
```diff
@@ -681,3 +681,25 @@ annotations:
...
+allowPrivilegeEscalation: true
+{{- else }}
+allowPrivilegeEscalation: false
+seccompProfile:
+    type: Unconfined
+capabilities:
+    drop:
+    - ALL
+    add:
+    {{- include "agent.capabilities" . | nindent 14 }}
```
mavimo:
```suggestion
    {{- include "agent.capabilities" . | nindent 4 }}
```

francesco-furlan (approved):

charts/agent/templates/_helpers.tpl
```diff
@@ -681,3 +681,25 @@ annotations:
...
+{{- else if eq "true" (include "agent.privileged" .) }}
+privileged: true
+runAsNonRoot: false
+runAsUser: 0
+readOnlyRootFilesystem: false
+allowPrivilegeEscalation: true
+{{- else }}
+allowPrivilegeEscalation: false
+seccompProfile:
+    type: Unconfined
```
mavimo:
```suggestion
  type: Unconfined
```

charts/agent/templates/_helpers.tpl
```diff
@@ -681,3 +681,25 @@ annotations:
...
+capabilities:
+    drop:
```
mavimo:
```suggestion
  drop:
```

charts/agent/templates/_helpers.tpl
```diff
@@ -681,3 +681,25 @@ annotations:
...
+    - ALL
+    add:
```
mavimo:
```suggestion
  add:
```

charts/agent/templates/_helpers.tpl
```diff
@@ -681,3 +681,25 @@ annotations:
 - SETUID
 - SETGID
 {{- end -}}
+
+
+{{- define "agent.securityContext" -}}
+{{- if .Values.customSecurityContext }}
+    {{- toYaml .Values.customSecurityContext -}}
```
mavimo:
```suggestion
  {{- toYaml .Values.customSecurityContext -}}
```

charts/agent/tests/security_context_test.yaml
mavimo, line 1: We already have some test related to security context in:
 * charts/agent/tests/conditional_flag_test.yaml
maybe we can move them all in the same file, WDYT?

mavimo (changes requested):

charts/agent/templates/_helpers.tpl
iurly: ok done

charts/agent/templates/_helpers.tpl
iurly: done

charts/agent/templates/_helpers.tpl
iurly: done

charts/agent/templates/_helpers.tpl
iurly: done

charts/agent/tests/security_context_test.yaml
iurly: it appeears to me the ones in conditional_flag_test actually test the consistency of the flag together with other values, wheres the ones here only focus on the securityContext. I would keep them separate.

mavimo (approved):

PR #2068, opened 2024-12-04
ci(rh-shield-operator): enhance operator release pipeline

aroberts87:
The old pipeline would simply build and push the operator and bundle images. This was a start, but left numerous manual steps to get the Operator itself certified. The changes in this PR add the following enhancements to the pipeline.

- Generate the Bundle content in the pipeline
  - Run `make bundle` in the pipeline as opposed to requiring it be run beforehand.
  - Since the pipeline guarantees the operator image itself will be built and pushed before the bundle is generated, we can set `USE_IMAGE_DIGESTS=true` when running `make bundle` to include the image checksums in the bundle. This is a requirement for certification.
  - The newly generated bundle content will be 'massaged' to include the annotations required for certification that are not created by `operator-sdk`.
4. Trigger `preflight` certification
3. Decouple the various builds and certification steps that aren't related

## Checklist


.github/workflows/release-rh-shield-operator.yaml
```diff
@@ -24,10 +40,103 @@ jobs:
...
+            sleep 10
+            PR_STATUS="$(gh pr view ${{ steps.open-pr.outputs.pull-request-number }} --json state -q .state)"
+          done'
+
+          if [[ "$PR_STATUS" != "MERGED" ]]; then
+            echo "PR was not merged in time. Check ${{ steps.open-pr.outputs.pull-request-url }} for more information."
+            exit 1
+          else
+            echo "PR was merged!"
+          fi
```
mavimo: What about using:
```sh
gh pr checks --watch  --json state ${{ steps.open-pr.outputs.pull-request-url }}
sleep 10
PR_STATUS="$(gh pr view ${{ steps.open-pr.outputs.pull-request-number }} --json state -q .state)"
echo "pr_status=$PR_STATUS" >> "$GITHUB_OUTPUT"
```
and then use other `if` condition in other steps to decide what is expected to happne?

```yaml
steps:
    - name: validate PR
      id: validate-pr
      # ...
    - name: Build and Push Bundle Image
      if: ${{ contains(steps.validate-pr.outputs.pr_status, 'MERGED') }}
      # ...
```

aroberts87: Maybe I'm not understanding correctly (😅), but the `gh pr checks` command seems to return as soon as the PR checks have completed and not when the PR is merged.  The need here is to wait until the merge has been completed so we do not build and push a bundle image that is based on content that hasn't actually been committed yet.

mavimo (approved):

PR #2128, opened 2025-01-23
chore(admission-controller): Bump Admission-controller to 0.16.8

lior-alafi-work:
## Checklist


mavimo:
@lior-alafi-work please integrate the values from `global.sysdig.*` as endpoints, see as example:
 * https://github.com/sysdiglabs/charts/blob/76c99599e97eb53315bf4e0baf3d01adebec8e61/charts/cluster-shield/templates/_helpers.tpl#L88-L92

and regions like:
 * https://github.com/sysdiglabs/charts/blob/76c99599e97eb53315bf4e0baf3d01adebec8e61/charts/agent/templates/_helpers.tpl#L348-L364

Thanks!

mavimo (changes requested):
see management for global values and regions

chen-shmilovich-sysdig:
@mavimo based on the admission controller helpers, it looks like we do use the `sysdig.global` values.
```
{{/*
Determine Secure endpoint based on provided region or .Values.sysdig.apiEndpoint
*/}}
{{- define "admissionController.apiEndpoint" -}}
    {{- if (or .Values.sysdig.apiEndpoint (eq .Values.global.sysdig.region "custom"))  -}}
        {{- required "A valid Sysdig API endpoint (.sysdig.apiEndpoint) is required" .Values.sysdig.apiEndpoint -}}
    {{- else if hasKey ((include "sysdig.regions" .) | fromYaml) .Values.global.sysdig.region }}
        {{- include "sysdig.secureApiEndpoint" . }}
    {{- end -}}
{{- end -}}
```

Should `.Values.global.sysdig.apiHost` be regarded or `.Values.global.sysdig.region` or both?
Thanks!

PR #2144, opened 2025-02-06
refactor(registry-scanner): rename scan on start job name [SSPROD-51637]

hayk99:

- Add FIPS image support 
- Fix startup job name 

## Checklist


charts/registry-scanner/templates/_helpers.tpl
```diff
@@ -105,11 +105,22 @@ Allow overriding registry and repository for air-gapped environments
...
 {{- end -}}
 
+{{- define "registry-scanner.inlineScanImage" -}}
+{{- if .Values.image.fips -}}
+    {{- .Values.config.scan.inlineScanImage -}} -fips
+{{- else -}}
+    {{- .Values.config.scan.inlineScanImage -}}
+{{- end -}}
+{{- end -}}
+
```
mech-pig, line 123: just a doubt: if there is an override of the image name, is the inlineScanImage going to remain the same as in `.Values.config.scan.inlineScanImage[-fips]`?

mech-pig (approved):
lgtm

charts/registry-scanner/templates/_helpers.tpl
hayk99: short answer, yes. You can customize the manager image and keep the inline scan with just the `-fips` prefix

Ideally, manger and worker images should be aligned. As of today, they are not, since inlineScanImage points to `:latest` unless it's explicitly requested with a different name.

An approach that we can take is to remove this parameter from the values and configure it based on the `image.name`, `image.tag`, etc values, which I think it's the best approach to keep both image names aligned. However, this will be a breaking change in the charts since we are removing a value, but I think it's still a good idea. Wdyt?

mech-pig: I wouldn't introduce a breaking change here as of now, so let's go ahead with this

PR #2148, opened 2025-02-10
chore (node-analyzer): add sslVerifyCertificate=false

radhikapc:
## Checklist

### DIFF
--- a/charts/node-analyzer/README.md
+++ b/charts/node-analyzer/README.md
@@ -40,7 +40,8 @@ helm install sysdig-agent --namespace sysdig-agent --create-namespace \
 --set nodeAnalyzer.secure.vulnerabilityManagement.newEngineOnly=true \
 --set global.kspm.deploy=true \
 --set nodeAnalyzer.nodeAnalyzer.benchmarkRunner.deploy=false \
---set nodeAnalyzer.nodeAnalyzer.hostScanner.deploy=true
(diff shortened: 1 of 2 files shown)


charts/node-analyzer/README.md
```diff
@@ -40,7 +40,8 @@ helm install sysdig-agent --namespace sysdig-agent --create-namespace \
 --set nodeAnalyzer.secure.vulnerabilityManagement.newEngineOnly=true \
 --set global.kspm.deploy=true \
 --set nodeAnalyzer.nodeAnalyzer.benchmarkRunner.deploy=false \
---set nodeAnalyzer.nodeAnalyzer.hostScanner.deploy=true
+--set nodeAnalyzer.nodeAnalyzer.hostScanner.deploy=true \
+--set nodeAnalyzer.nodeAnalyzer.sslVerifyCertificate=false \
```
airadier, line 44: Shouldn't the `sslVerifyCertificate` be added to the table below, and not to the example?

But the value is there. Is it just missing the default value:
image

radhikapc: What is the default value? I thought we need to add SSL verification to the example

airadier: Adding to the example is ok. But the default value is missing. The default value AFAIK is "true" (verify certificate by default), or maybe it is empty achieving the same effect

francesco-furlan:
@radhikapc is this still relevant?

PR #2149, opened 2025-02-10
fix(registry-scanner): handle error for fips image

hayk99:
## Checklist

### DIFF
--- a/charts/registry-scanner/templates/configmap.yaml
+++ b/charts/registry-scanner/templates/configmap.yaml
@@ -82,7 +82,7 @@ data:
           limits:
             memory: {{ .Values.config.scan.jobs.resources.limits.memory }}
         temporaryVolumeSizeLimit: {{ .Values.config.scan.jobs.temporaryVolumeSizeLimit }}
-      {{- if .Values.config.scan.inlineScanImage }}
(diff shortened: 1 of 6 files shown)


charts/registry-scanner/templates/configmap.yaml
```diff
@@ -82,7 +82,7 @@ data:
           limits:
             memory: {{ .Values.config.scan.jobs.resources.limits.memory }}
         temporaryVolumeSizeLimit: {{ .Values.config.scan.jobs.temporaryVolumeSizeLimit }}
-      {{- if .Values.config.scan.inlineScanImage }}
+      {{- if  or .Values.config.scan.inlineScanImage .Values.image.fips }}
```
dbonino, line 85: I do not see the point of adding `.Values.image.fips` in the value check if it is not used in the `if` body, can you please clarify this?

hayk99: In order to force the workers to use the fips images, we have to pass the `inlineScanImage:` parameter. Otherwise, the workers will use the image tagged in the build process see [here](https://github.com/draios/registry-scanner/blob/master/build/k8sjob/Dockerfile#L23)
This is a way to force the workers to use the fips image in case it's requested

dbonino (approved):
LGTM

PR #2184, opened 2025-03-19
feat(shield): add support for windows host-shield

francesco-furlan:
## Checklist


charts/shield/values.yaml
```diff
@@ -152,6 +152,79 @@ features:
...
       enabled: false
 
+host_windows:
+  # Enable the host shield for Windows
+  enabled: false
+  # Additional settings to be passed to the host shield (overrides the helm generated settings)
+  additional_settings:
+    monitoring_port: 8081
+  # The agent-runtime additional settings (overrides dragent.yaml)
+  runtime_additional_settings: {}
```
francesco-furlan: let's rename it to shield_additional_settings and use this for `host-shield.yaml` instead of `dragent.yaml`

charts/shield/templates/_helpers.tpl
```diff
@@ -23,6 +23,22 @@ If release name contains chart name it will be used as a full name.
...
+    {{- else }}
+        {{- $name := (include "shield.name" .) }}
+        {{- if contains $name .Release.Name }}
+            {{- printf "%s-%s" "windows" .Release.Name | trunc 63 | trimSuffix "-" }}
+        {{- else }}
+            {{- printf "%s-%s-%s" "windows" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+        {{- end }}
+    {{- end }}
+{{- end }}
+
```
AlbertoBarba: Not needed...u can simply use the `shield.fullname` passing `host-windows`.
It will also help to keep consistency between object naming`<RELEASE-NAME>-<COMPONENT>`

charts/shield/templates/host/_helpers.tpl
```diff
@@ -7,8 +7,32 @@ If release name contains chart name it will be used as a full name.
     {{- printf "%s-host" (include "shield.fullname" . | trunc 57 | trimSuffix "-") }}
 {{- end }}
 
+{{- define "host.windows.fullname" -}}
+    {{- printf "%s-host" (include "shield.windows.fullname" . | trunc 57 | trimSuffix "-") }}
```
AlbertoBarba:
```suggestion
    {{- printf "%s-host-windows" (include "shield.fullname" . | trunc 57 | trimSuffix "-") }}
```

charts/shield/templates/host/_windows_annotations.tpl
```diff
@@ -0,0 +1,14 @@
...
+  {{- with $workloadAnnotations -}}
+    {{- . | toYaml -}}
+  {{- end -}}
+{{- end -}}
+
+{{- define "host.windows.pod_annotations" -}}
+  {{- $podAnnotations := merge (dict) .Values.pod_annotations .Values.host_windows.pod_annotations -}}
+  {{- if (include "common.cluster_type.is_gke_autopilot" . ) -}}
+    {{- $_ := set $podAnnotations "autopilot.gke.io/no-connect" "true" -}}
+  {{- end -}}
```
AlbertoBarba: Not sure GKE Autopilot supports Windows 🤔

charts/shield/templates/host/_windows_configmap_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+  {{- end }}
+  {{- $_ := set $config "tags" (join "," $tagList) -}}
+{{- end -}}
+{{- if (include "common.proxy.enabled" . ) }}
+{{- $config := merge $config (dict "http_proxy" (include "host.dragent_proxy_config" . | fromYaml)) }}
+{{- end }}
+{{- $override := (include "host.windows.runtime_config_override" .) | fromYaml }}
+{{- $finalConfig := mergeOverwrite $config $override }}
+{{- $finalConfig | toYaml }}
+{{- end }}
```
AlbertoBarba: nit: might be useful to indent whenever it's possible to increase readability

AlbertoBarba (changes requested):

charts/shield/templates/_helpers.tpl
francesco-furlan: I did it on purpose to have `windows` as a prefix, otherwise we could end up with the same name when release name + component is too long

charts/shield/templates/host/_windows_annotations.tpl
francesco-furlan: makes sense, I will remove it

charts/shield/templates/host/_windows_configmap_helpers.tpl
francesco-furlan: agree

charts/shield/templates/_helpers.tpl
francesco-furlan: @mavimo @aroberts87 we were discussing that to keep a sort of naming standard it could make sense to have somthing like `<RELEASE-NAME>-windows-<COMPONENT>`. Do you agree?

francesco-furlan: refactored to `<RELEASE-NAME>-<COMPONENT>-windows` ✅

AlbertoBarba (approved):

PR #2212, opened 2025-04-08
chore(shield): use volumeMount instead of env for rapid-response pwd

francesco-furlan:

Using `rapid-response` password secret as volume instead of env.

NOTE: as of now we need to use `subPath` volumeMounts since access-key is under the standard `/opt/draios/etc/kubernetes/secrets/` while rapid-response password has an additional `rapid-response` folder.

## Checklist


francesco-furlan:
We decided to postpone this refactor when agent will support different paths for secrets, so we can get rid of the `subPath` approach (which requires a re-deployment once the secrets changes, and we want to avoid that)

PR #2213, opened 2025-04-08
feat(shield): Allow configurable DNS Policy for Host Shield

aroberts87:
SSIA

## Checklist


francesco-furlan (approved):
LGTM
just one small thing around the "default" function I recall we had issues in the past, but maybe it was related to some old helm versions that we don't support anymore

charts/shield/values.yaml
```diff
@@ -362,6 +362,8 @@ host:
   volumes: []
   # The custom volume mounts for the host shield
   volume_mounts: []
+  # The dns policy for the host shield
+  dns_policy: ClusterFirstWithHostNet
```
mavimo: We do not have this as default in the cluster section, I'll prefer to be consistent here.

aroberts87: Let me know if the changes in https://github.com/sysdiglabs/charts/pull/2213/commits/71e0111657eda8d416b4044a244653d0d2c279b8 address your concern here. 👍🏻

mavimo: LGTM

mavimo (approved):

PR #2216, opened 2025-04-10
feat(shield): improve regions handling logic to allow override region specific settings without setting region to custom

maratsal:
improve regions handling logic to allow override region specific settings without setting region to custom. Currently if we want to use alternative collector settings we have to use custom region and specify each of the following values:
```
sysdig_endpoint:
  region: custom
  api_url: https://www.example.com
  collector:
    host: example.com
    port: 6443
```
Many customers want to keep using default region settings and update only collector endpoints and use config like below:
```
sysdig_endpoint:
  region: us2
  collector:
    host: alt-example.com
    port: 443
```

This PR is addressing that functionality.

## Checklist

### DIFF
--- a/charts/shield/templates/common/_regions.tpl
+++ b/charts/shield/templates/common/_regions.tpl
@@ -180,8 +180,8 @@
 
 {{- define "common.collector_endpoint" }}
   {{- $regions := fromYaml (include "common.regions" .) }}
-  {{- if ne "custom" .Values.sysdig_endpoint.region }}
-    {{- get (get $regions .Values.sysdig_endpoint.region) "collector_endpoint"  }}
+  {{- if not .Values.sysdig_endpoint.collector.host }}
+    {{- get (get $regions .Values.sysdig_endpoint.region) "collector_endpoint" }}
   {{- else }}
     {{- .Values.sysdig_endpoint.collector.host }}
   {{- end }}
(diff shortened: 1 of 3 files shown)


charts/shield/templates/common/_regions.tpl
```diff
@@ -180,16 +180,16 @@
 
 {{- define "common.collector_endpoint" }}
   {{- $regions := fromYaml (include "common.regions" .) }}
-  {{- if ne "custom" .Values.sysdig_endpoint.region }}
-    {{- get (get $regions .Values.sysdig_endpoint.region) "collector_endpoint"  }}
+  {{- if and (ne "custom" .Values.sysdig_endpoint.region) (not .Values.sysdig_endpoint.collector.host) }}
```
aroberts87: @maratsal - If I understand the intent of this PR I think what the intent of this change is, is to use `sysdig_endpoint.collector.host` any time it is specified, correct? If that's true, we could simplify the boolean logic here.

If my understanding is correct, you could apply the simplification to the other areas you've modified as well.

cc @AlbertoBarba

maratsal: Maybe, if you can share what you are thinking of - my brain always comes back to couple of checks anyways :D

aroberts87: Well, if I'm not mistaken, this change is intended to use `sysdig_endpoint.collector.host` whether the region field is `custom` or some valid region, correct? If that's the case, you can simplify the checks here to simply see if `sysdig_endpoint.collector.host` is non-null.

maratsal: updated logic.

aroberts87 (approved):

PR #2251, opened 2025-05-14
feat(kspm-collector): grant k8s lease permissions if transport is http [SSPROD-56744]

mech-pig:


mech-pig:
superseded by https://github.com/sysdiglabs/charts/pull/2255

PR #2325, opened 2025-07-19
Modify the shield helm chart to work with cert-manager

yoderme:
The certificates in the shield helm charts do not play nicely with argocd. Since argocd will essentially call "helm template" to acquire the chart's yaml, the "lookup" function will never work, and the certificate generation will create a different certificate each time.

It's possible to tell ArgoCD to ignore this ever-changing certificate, but this also means that the certificate will never be rotated.

The most graceful solution to this is to integrate with the widely-used cert-manager project. See
https://github.com/cert-manager/cert-manager

The idea is that you can create a certificate chain using cert-manager, and then refer to those certificate in the helm chart instead of asking the helm chart to generate them.

For our ppurposes there are two interesting cert-manager resources. The "Issuer" is a resource that is used to sign a certificate. A "Certificate" is actually a request to create an x509 certificate. It emits a k8s Secret containing the actual TLS certificate details.

See:
* https://cert-manager.io/docs/concepts/issuer/
* https://cert-manager.io/docs/usage/certificate/

Imagine that the customer has the following setup:
* A) a "self-signed" root Issuer
* B) a "CA" Certificate representing the root (using A above). It will be long-lived, like all root CAs are.
* C) a "root cert" Secret containing the actual x509 root CA cert (created by cert-manager as a result of B)
* D) a "CA" Issuer that uses the Secret from C above

At this point there is a long-lived root x509 certificate as well as an Issuer that will create certificates signed by that root.

At this point, the shield helm chart can include a Certificate resource that refers to the Issuer from step D above. This will generate a Secret with the actual x509 cert. This secret can be used by the shield Deployment for the TLS webhook port.

Additionally, the validating webhook can refer to the CA Certificate resource from step B above. It uses an annotation that tells cert-manger to edit the webhook and insert the appropriate caBundle information.

### DIFF
--- a/charts/shield/templates/cluster/_tls.tpl
+++ b/charts/shield/templates/cluster/_tls.tpl
@@ -9,12 +9,54 @@
 {{- end }}
 
 {{- define "cluster.tls_certificates.secret_name" -}}
-  {{- if .Values.cluster.tls_certificates.create -}}
+  {{- if or .Values.cluster.tls_certificates.create .Values.cluster.tls_certificates.cert_manager.enabled -}}
     {{- include "cluster.fullname" . }}-tls-certificates
   {{- else if .Values.cluster.tls_certificates.secret_name -}}
     {{- .Values.cluster.tls_certificates.secret_name -}}
   {{- else -}}
     {{- fail "TLS Certificates must be generated or specified with the current configuration" -}}
+  {{- end -}}
+{{- end }}
+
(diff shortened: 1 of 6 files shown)


francesco-furlan:
Handled in this PR https://github.com/sysdiglabs/charts/pull/2411 thanks @yoderme for your contribution! 🙌

PR #2416, opened 2025-10-24
fix(shield): set seLinuxOptions to control_t for unprivileged host shield bottlerocket deployment

therealbobo:

This PR adds conditional SELinux configuration support for Bottlerocket clusters to enable response actions in unprivileged host shield deployments.

## Checklist


charts/shield/templates/host/_helpers.tpl
```diff
@@ -201,6 +201,10 @@ capabilities:
 allowPrivilegeEscalation: false
 seccompProfile:
   type: Unconfined
+{{- if (include "common.cluster_type.is_bottlerocket" .) }}
```
francesco-furlan: let's add also the check if response actions is enabled

francesco-furlan (approved):

PR #2605, opened 2026-04-29
fix(shield): correct response_actions ClusterRole RBAC scoping

### LINKED ISSUE #2603: [BUG] `cluster.rbac.create: false` causes `helm template` to fail with `SimpleHead` unmarshal error in `templates/cluster/clusterrole.yaml`
## Summary

When installing the `shield` chart with `cluster.rbac.create: false`, Helm fails to render `templates/cluster/clusterrole.yaml`. The rendered output contains YAML rule arrays (`- apiGroups: ...`) without the corresponding `apiVersion / kind / metadata / rules:` document head, which Helm cannot parse as a valid manifest.

The same option used to work in earlier chart versions and is documented in the values file:

```yaml
# values.yaml (excerpt)
service_account_name:
  # The name of the service account for the cluster shield
  # (if create is set to false, this will be used as the name
  #  of the existing service account)
```

So this looks like an unintended regression in `templates/cluster/clusterrole.yaml`, not a documentation issue.

The same problem does **not** occur in `templates/host/clusterrole.yaml`, which wraps every rule block inside the outer `{{- if .Values.host.rbac.create }} ... {{- end }}` correctly.

## Affected versions

- Chart: `sysdig/shield` **1.36.0** (cluster-shield image **1.22.0**)
- Reproduces against `main` branch HEAD as of the date of this report.

## Environment

- Helm: `v3.16.2`
- This is a Helm template-time error, so it is independent of the target Kubernetes cluster.

## Steps to reproduce

1. Save the following minimal `values.yaml`:

   ```yaml
   cluster_config:
     name: test-cluster
   sysdig_endpoint:
     region: us1
     access_key: dummy

   cluster:
     rbac:
       create: false
       service_account_name: my-existing-sa
   ```

2. Run:

   ```bash
   helm template test ./charts/shield -f values.yaml
   ```

3. Observe the error:

   ```
   Error: YAML parse error on shield/templates/cluster/clusterrole.yaml:
   error unmarshaling JSON: while decoding JSON:
   json: cannot unmarshal array into Go value of type releaseutil.SimpleHead
   ```

4. Inspect the rendered output with `--debug`:

   ```bash
   helm template test ./charts/shield -f values.yaml \
     --show-only templates/cluster/clusterrole.yaml --debug
   ```

   The rendered file contains `- apiGroups: ...` array entries with no `apiVersion / kind / metadata / rules:` header above them.

## Expected behavior

When `cluster.rbac.create: false`:

- `templates/cluster/clusterrole.yaml` should render to an empty manifest (no document emitted), and
- `helm template` / `helm install` should succeed.

This is what `templates/host/clusterrole.yaml` already does correctly, and what the `cluster/role.yaml`, `cluster/rolebinding.yaml`, `cluster/clusterrolebinding.yaml`, and `cluster/serviceaccount.yaml` templates do correctly.

## Actual behavior

`helm template` and `helm install` fail with `cannot unmarshal array into Go value of type releaseutil.SimpleHead`.

The rendered output ends with stray rule entries that have no enclosing document, e.g.:

```yaml
- apiGroups:
    - ""
  resources:
    - pods/log
  verbs:
    - get
- apiGroups:
    - batch
  resources:
    - jobs
  verbs:
    - get
# ... (more bare rule entries) ...
```

## Root cause

In [`[templates/cluster/clusterrole.yaml](https://github.com/sysdiglabs/charts/blob/main/charts/shield/templates/cluster/clusterrole.yaml)`](https://github.com/sysdiglabs/charts/blob/main/charts/shield/templates/cluster/clusterrole.yaml), the outer `{{- if .Values.cluster.rbac.create }} ... {{- end }}` wrapper closes at line 311. The eight `response_actions.*` rule blocks (lines 313–457) are placed **outside** that wrapper:

```gotemplate
1     {{- if .Values.cluster.rbac.create }}        # outer if
        apiVersion / kind / metadata / rules:
        ├─ audit_enabled              (closes line 37)    OK (inside)
        ├─ container_vuln_mgmt        (closes line 82)    OK
        ├─ posture                    (closes line 192)   OK
        ├─ admission_control          (closes line 204)   OK
        └─ kubernetes_metadata + network_security (310)   OK
311   {{- end }}                                          # outer if closes here

313   {{- if eq "true" (include "cluster.response_actions_enabled" .) }}                     OUTSIDE
323   {{- if eq "true" (include "cluster.response_actions.rollout_restart.enabled" .) }}     OUTSIDE
335   {{- if eq "true" (include "cluster.response_actions.delete_pod.enabled" .) }}          OUTSIDE
345   {{- if eq "true" (include "cluster.response_actions.isolate_network.enabled" .) }}     OUTSIDE
370   {{- if eq "true" (include "cluster.response_actions.delete_network_policy.enabled" .) }} OUTSIDE
380   {{- if eq "true" (include "cluster.response_actions.get_logs.enabled" .) }}            OUTSIDE
413   {{- if eq "true" (include "cluster.response_actions.volume_snapshot.enabled" .) }}     OUTSIDE
447   {{- if eq "true" (include "cluster.response_actions.delete_volume_snapshot.enabled" .) }} OUTSIDE
```

Combined with the helper definition in [`[templates/cluster/_config.tpl](https://github.com/sysdiglabs/charts/blob/main/charts/shield/templates/cluster/_config.tpl)`](https://github.com/sysdiglabs/charts/blob/main/charts/shield/templates/cluster/_config.tpl):

```gotemplate
{{- define "cluster.response_actions.is_enabled" -}}
    {{- $action := .Action }}
    {{- $ctx := .Context }}
    {{- with $ctx.Values.features.respond.response_actions -}}
        {{- $entry := index . $action }}
        {{- if and $entry (eq $entry.trigger "none") -}}
            false
        {{- else -}}
            true        # <-- defaults to true unless trigger == "none"
        {{- end -}}
    {{- else -}}
        true
    {{- end -}}
{{- end -}}
```

…each individual action helper (`rollout_restart.enabled`, `delete_pod.enabled`, `isolate_network.enabled`, `delete_network_policy.enabled`, `get_logs.enabled`, `volume_snapshot.enabled`, `delete_volume_snapshot.enabled`) returns the literal string `"true"` by default, because the user almost never sets `trigger: "none"` explicitly under `features.respond.response_actions.<action>`.

So when `cluster.rbac.create: false`:

1. The outer `if` drops the `apiVersion / kind / metadata / rules:` header (lines 1–311 produce nothing).
2. The 7 individual `response_actions.*` rule snippets at lines 323–457 still render at the top level, because their helpers default to `"true"`.
3. The result is a YAML document that begins with `- apiGroups: ...` (an array), which Helm tries to parse as a `releaseutil.SimpleHead` (a map containing `apiVersion / kind / metadata`), and fails.

(The block at line 313, `cluster.response_actions_enabled`, depends on `features.respond.response_actions.enabled` and correctly returns an empty string by default, so it does not contribute to the bug — only the seven individual action blocks do.)

## Suggested fix

Move the seven individual `response_actions.*` rule blocks (lines 323–457) **inside** the outer `{{- if .Values.cluster.rbac.create }} ... {{- end }}` wrapper, matching the pattern already used in `templates/host/clusterrole.yaml` and the rest of `templates/cluster/`.

Conceptually:

```gotemplate
{{- if .Values.cluster.rbac.create }}
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  ...
rules:
  {{- /* existing audit/posture/CVM/admission/metadata/network_security blocks */ -}}

  {{- /* existing response_actions blocks, indented with two spaces */ -}}
  {{- if eq "true" (include "cluster.response_actions_enabled" .) }}
  - apiGroups:
      - ""
    resources:
      - events
    ...
  {{- end }}
  {{- if eq "true" (include "cluster.response_actions.rollout_restart.enabled" .) }}
  - apiGroups: ["apps"]
    ...
  {{- end }}
  {{- /* ... and the other 5 response_actions blocks ... */ -}}
{{- end }}
```

Optionally, the indentation of the `response_actions` rule entries could be aligned (current entries have `- apiGroups:` flush left, while the rules above use `  - apiGroups:` indented two spaces — both are valid YAML, but consistency would help diffing/review).

## Workarounds (verified)

Both of these were verified locally with `helm template test ./charts/shield -f workaround.yaml` against the `main` branch.

### Workaround 1 — preferred: let the chart create the RBAC

Set `cluster.rbac.create: true` (the default) and just specify `service_account_name`. The values.yaml comment confirms `service_account_name` is honored either way:

```yaml
cluster:
  rbac:
    create: true
    service_account_name: my-existing-sa   # used as the name of the SA the chart creates
host:
  rbac:
    create: true
    service_account_name: my-existing-sa
```

This avoids the buggy code path entirely.

### Workaround 2 — keep `rbac.create: false`, suppress every response action

If RBAC must be managed externally:

```yaml
cluster:
  rbac:
    create: false
    service_account_name: my-existing-sa

features:
  respond:
    response_actions:
      enabled: false
      rollout_restart:        { trigger: "none" }
      delete_pod:             { trigger: "none" }
      isolate_network:        { trigger: "none" }
      delete_network_policy:  { trigger: "none" }
      get_logs:               { trigger: "none" }
      volume_snapshot:        { trigger: "none" }
      delete_volume_snapshot: { trigger: "none" }
```

This forces every individual `response_actions.*.is_enabled` helper to return `false`, so the offending blocks are skipped and `clusterrole.yaml` renders to an empty manifest as expected.

This workaround is fragile: if a new response action is added in a future release without a corresponding `trigger: "none"` override in the user's values, the bug will silently come back.

## Why I am opening this issue

The fix in the chart itself (moving the 7 blocks inside the outer `if`) is small and isolated, and would let users keep the `rbac.create: false` deployment pattern that the values.yaml documentation already advertises. Happy to open a PR if that helps.

---

### Reference

- Affected file: [`[charts/shield/templates/cluster/clusterrole.yaml](https://github.com/sysdiglabs/charts/blob/main/charts/shield/templates/cluster/clusterrole.yaml)`](https://github.com/sysdiglabs/charts/blob/main/charts/shield/templates/cluster/clusterrole.yaml)
- Helper definitions: [`[charts/shield/templates/cluster/_config.tpl](https://github.com/sysdiglabs/charts/blob/main/charts/shield/templates/cluster/_config.tpl)`](https://github.com/sysdiglabs/charts/blob/main/charts/shield/templates/cluster/_config.tpl)
- Comparison (working pattern): [`[charts/shield/templates/host/clusterrole.yaml](https://github.com/sysdiglabs/charts/blob/main/charts/shield/templates/host/clusterrole.yaml)`](https://github.com/sysdiglabs/charts/blob/main/charts/shield/templates/host/clusterrole.yaml)
- Helm `releaseutil.SimpleHead` for context: <https://pkg.go.dev/helm.sh/helm/v3/pkg/releaseutil>

francesco-furlan:
## Summary

Fixes two related defects in the cluster-shield ClusterRole template that caused `response_actions` RBAC to render incorrectly.

### Bug 1 — per-action gates ignored the master `enabled` flag

`cluster.response_actions.<action>.is_enabled` (in `templates/cluster/_config.tpl`) only short-circuited when an individual action's `trigger` was explicitly `"none"` — it never consulted `features.respond.response_actions.enabled`. With chart defaults (master flag `false`), every per-action gate still resolved to `"true"` and the cluster-shield ClusterRole was granted `delete` on `pods`, `networkpolicies`, and `volumesnapshots`, plus the `isolate_network` / `rollout_restart` / `get_logs` / `volume_snapshot` rules. Real least-privilege violation.

### Bug 2 — response_actions blocks rendered outside the `cluster.rbac.create` wrapper

The seven `response_actions.*` rule blocks in `templates/cluster/clusterrole.yaml` lived **outside** the outer `{{- if .Values.cluster.rbac.create }} ... {{- end }}` guard. With `cluster.rbac.create: false`, the document head (`apiVersion / kind / metadata / rules:`) was correctly suppressed but the per-action rule snippets still rendered, producing a top-level YAML array that Helm could not parse:

```
Error: YAML parse error on shield/templates/cluster/clusterrole.yaml:
error unmarshaling JSON: while decoding JSON:
json: cannot unmarshal array into Go value of type util.SimpleHead
```

Reported by @EdwardArchive in #2603.

## Fix

- `templates/cluster/_config.tpl`: `is_enabled` returns `false` early when the master `features.respond.response_actions.enabled` flag is falsy. Per-action `trigger: "none"` overrides remain effective when the master flag is on.
- `templates/cluster/clusterrole.yaml`: moved the `{{- end }}` of the `cluster.rbac.create` wrapper past the seven `response_actions.*` blocks so they live inside it (matching the working pattern in `templates/host/clusterrole.yaml`).
- `Chart.yaml`: bumped to `1.36.1`.

## Behavior matrix

| `response_actions.enabled` | `<action>.trigger` | Before | After |
|---|---|---|---|
| `false` (default) | unset | rules **leaked** into ClusterRole | rules absent |
| `false` | `"none"` | rules absent | rules absent |
| `true` | unset | rules present | rules present |
| `true` | `"none"` | rules absent | rules absent |

| `cluster.rbac.create` | `response_actions.enabled` | Before | After |
|---|---|---|---|
| `true` (default) | any | renders ClusterRole | renders ClusterRole |
| `false` | `false` (default) | **`helm template` error** | renders no document |
| `false` | `true` | **`helm template` error** | renders no document |

## Tests

`charts/shield/tests/cluster/clusterrole_test.yaml`:

- **`response_actions disabled by default does not leak per-action RBAC`** — asserts `pods: delete,get`, `networkpolicies: get,delete`, `volumesnapshots: delete,get,watch,patch`, and `pods/log: get` are absent under chart defaults.
- **`cluster.rbac.create false renders no ClusterRole`** — `hasDocuments: count: 0`.
- **`cluster.rbac.create false with response_actions enabled still renders no ClusterRole`** — covers the original failure mode from #2603.
- **`response_actions enabled with delete_pod trigger none suppresses only delete_pod rule`** — verifies fine-grained per-action overrides still work.

`helm unittest --strict -f "tests/**/*_test.yaml" charts/shield`: **463 / 463 pass** (459 baseline + 4 new).

## Test plan

- [x] `helm unittest` passes (463/463)
- [x] `helm template` with `cluster.rbac.create: false` no longer errors
- [x] Default render (no overrides) emits no response-action rules in cluster-shield ClusterRole
- [x] `features.respond.response_actions.enabled: true` still emits all per-action rules
- [x] `enabled: true` + per-action `trigger: "none"` suppresses only that rule
- [ ] CI lint + helm-unit-test workflows

Closes #2603

### DIFF
--- a/charts/shield/Chart.yaml
+++ b/charts/shield/Chart.yaml
@@ -13,5 +13,5 @@ maintainers:
   - name: mavimo
     email: marcovito.moscaritolo@sysdig.com
 type: application
-version: 1.36.0
+version: 1.36.1
 appVersion: "1.0.0"
--- a/charts/shield/templates/cluster/_config.tpl
+++ b/charts/shield/templates/cluster/_config.tpl
@@ -224,21 +224,28 @@
 {{- end }}
 
 {{/*
-Generic helper: checks if .Values.features.respond.response_actions.<action>.trigger == "all"
+Generic helper: returns "true" when the response_actions feature is enabled at
+the master level (.Values.features.respond.response_actions.enabled) AND the
+specific per-action trigger is not explicitly set to "none". Returns "false"
+otherwise.
 Usage: {{ include "cluster.response_actions.is_enabled" (dict "Action" "delete_pod" "Context" .) }}
 */}}
 {{- define "cluster.response_actions.is_enabled" -}}
     {{- $action := .Action }}
     {{- $ctx := .Context }}
-    {{- with $ctx.Values.features.respond.response_actions -}}
-        {{- $entry := index . $action }}
-        {{- if and $entry (eq $entry.trigger "none") -}}
-            false
+    {{- if eq "true" (include "cluster.response_actions_enabled" $ctx) -}}
+        {{- with $ctx.Values.features.respond.response_actions -}}
+            {{- $entry := index . $action }}
+            {{- if and $entry (eq $entry.trigger "none") -}}
+                false
+            {{- else -}}
+                true
+            {{- end -}}
         {{- else -}}
             true
         {{- end -}}
     {{- else -}}
-        true
+        false
     {{- end -}}
 {{- end -}}
 
--- a/charts/shield/templates/cluster/clusterrole.yaml
+++ b/charts/shield/templates/cluster/clusterrole.yaml
@@ -308,7 +308,6 @@ rules:
     - list
     - watch
 {{- end }}
-{{- end }}
 
 {{- if eq "true" (include "cluster.response_actions_enabled" .) }}
 - apiGroups:
@@ -454,4 +453,5 @@ rules:
     - get
     - watch
     - patch # needed to remove finalizers, which could prevent deletion
+{{- end }}
 {{- end }}
--- a/charts/shield/tests/cluster/clusterrole_test.yaml
+++ b/charts/shield/tests/cluster/clusterrole_test.yaml
@@ -837,3 +837,97 @@ tests:
             verbs:
               - create
               - patch
+
+  - it: response_actions disabled by default does not leak per-action RBAC
+    asserts:
+      - containsDocument:
+          kind: ClusterRole
+          apiVersion: rbac.authorization.k8s.io/v1
+          name: release-name-shield-cluster
+      - notContains:
+          path: rules
+          content:
+            apiGroups:
+              - ""
+            resources:
+              - pods
+            verbs:
+              - delete
+              - get
+      - notContains:
+          path: rules
+          content:
+            apiGroups:
+              - networking.k8s.io
+            resources:
+              - networkpolicies
+            verbs:
+              - get
+              - delete
+      - notContains:
+          path: rules
+          content:
+            apiGroups:
+              - snapshot.storage.k8s.io
+            resources:
+              - volumesnapshots
+            verbs:
+              - delete
+              - get
+              - watch
+              - patch
+      - notContains:
+          path: rules
+          content:
+            apiGroups:
+              - ""
+            resources:
+              - pods/log
+            verbs:
+              - get
+
+  - it: response_actions enabled with delete_pod trigger none suppresses only delete_pod rule
+    set:
+      features:
+        respond:
+          response_actions:
+            enabled: true
+            delete_pod:
+              trigger: none
+    asserts:
+      - containsDocument:
+          kind: ClusterRole
+          apiVersion: rbac.authorization.k8s.io/v1
+          name: release-name-shield-cluster
+      - notContains:
+          path: rules
+          content:
+            apiGroups:
+              - ""
+            resources:
+              - pods
+            verbs:
+              - delete
+              - get
+      - contains:
+          path: rules
+          content:
+            apiGroups:
+              - networking.k8s.io
+            resources:
+              - networkpolicies
+            verbs:
+              - get
+              - delete
+      - contains:
+          path: rules
+          content:
+            apiGroups:
+              - snapshot.storage.k8s.io
+            resources:
+              - volumesnapshots
+            verbs:
+              - delete
+              - get
+              - watch
+              - patch
--- a/charts/shield/tests/cluster/clusterrole_test.yaml
+++ b/charts/shield/tests/cluster/clusterrole_test.yaml
@@ -886,6 +886,28 @@ tests:
             verbs:
               - get
 
+  - it: cluster.rbac.create false renders no ClusterRole
+    set:
+      cluster:
+        rbac:
+          create: false
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  - it: cluster.rbac.create false with response_actions enabled still renders no ClusterRole
+    set:
+      cluster:
+        rbac:
+          create: false
+      features:
+        respond:
+          response_actions:
+            enabled: true
+    asserts:
+      - hasDocuments:
+          count: 0
+
   - it: response_actions enabled with delete_pod trigger none suppresses only delete_pod rule
     set:
       features:


EdwardArchive:
Thank you for hard work @francesco-furlan !

francesco-furlan:
> Thank you for hard work @francesco-furlan !

Thank you for raising the issue @EdwardArchive !  🚀

AlbertoBarba (approved):

PR #2528, opened 2026-02-16
feat(shield): add gke-allowlist-waiter job for gke-autopilot clusters

francesco-furlan:
## Checklist

- [ ] Title of the PR starts with type and scope, (e.g. `feat(agent,node-analyzer,sysdig-deploy):`)
- [ ] Chart Version bumped for the respective charts
- [ ] Variables are documented in the README.md (or README.tpl in some charts)
- [ ] Check GithubAction checks (like lint) to avoid merge-check stoppers
- [ ] All test files are added in the tests folder of their respective chart and have a "_test" suffix

PR #2606, opened 2026-04-29
feat(shield): pre-install AllowlistSynchronizer waiter for GKE Autopilot

francesco-furlan:
## What this PR does / why we need it

On GKE Autopilot, the host-shield DaemonSet can roll out before the cluster's `AllowlistSynchronizer/sysdig-agent-allowlist-synchronizer` reaches `Ready`. When that happens Autopilot blocks the host-shield pods until the allowlist matches them, and `helm install` / `helm upgrade` (or a Flux `HelmRelease` reconciliation) returns success while the DaemonSet is still wedged. There is no synchronization point in the chart today — the only existing Helm hook is the `AllowlistSynchronizer` resource itself, with no ordering relationship to the DaemonSet.

This PR adds a `pre-install` / `pre-upgrade` hook Job that `kubectl wait`s up to 120s for the AllowlistSynchronizer to be Ready before the DaemonSet is rolled out. It is **opt-in** and gated behind a new flag, so non-Autopilot installs are entirely unaffected.

Revives the approach from the stale, auto-closed [PR #2528](https://github.com/sysdiglabs/charts/pull/2528) and tightens it: gating flag, configurable+pinned image with its own pull secrets, README/values docs, version bump, and unit-test coverage.

## What's in the chart

Four new templates under `charts/shield/templates/host/`, all gated by `common.cluster_type.is_gke_autopilot` AND the new `gke_autopilot.allowlist_waiter.enabled` flag:

- `gke-allowlist-waiter-serviceaccount.yaml` (hook-weight `-5`)
- `gke-allowlist-waiter-clusterrole.yaml` (hook-weight `-5`) — limited to `auto.gke.io/allowlistsynchronizers` (`get,list,watch`)
- `gke-allowlist-waiter-clusterrolebinding.yaml` (hook-weight `-5`)
- `gke-allowlist-waiter-job.yaml` (hook-weight `5`, `hook-delete-policy: before-hook-creation,hook-succeeded`)

One existing template touched:

- `gke-allowlist-synchronizer.yaml` — added `helm.sh/hook-delete-policy: before-hook-creation`. Without this, the second `helm upgrade` would fail with `AlreadyExists` because the resource (a `pre-install,pre-upgrade` hook with no delete-policy until now) was never recycled. The waiter exists precisely to handle the brief NotReady transition this introduces — the two changes are complementary.

Hook ordering: SA / CR / CRB → existing AllowlistSynchronizer (default weight `0`) → waiter Job → non-hook resources (including the DaemonSet).

The Job runs the pinned `quay.io/sysdig/kubectl:1.34.3-1.6.21` image. Image pull secrets are configured independently from the host-shield image, since the kubectl image lives in a different repo and may need different credentials. The waiter Pod template intentionally uses `shield.labels` (not `host.labels`) so it does **not** carry `cloud.google.com/matching-allowlist` — otherwise GKE Warden would block the waiter Pod on the very allowlist it is waiting to load. A unit-test assertion guards this.

> Sweeping the rest of the repo to the maintained `sysdig/kubectl` image (other 4 charts + `.github/updatecli.d/config-update-kubectl-image.yaml`) is intentionally **out of scope** here — follow-up PR.

## Configuration

All GKE-Autopilot-related knobs are consolidated under a new top-level `gke_autopilot` section:

```yaml
gke_autopilot:
  allowlist_version:           # replaces top-level gke_autopilot_allowlist (still honored as a fallback)
  allowlist_waiter:
    enabled: false             # default off — opt-in
    timeout: 120s
    service_account_name:      # override; defaults to <host-fullname>-allowlist-waiter
    image:
      registry: quay.io
      repository: sysdig/kubectl
      tag: 1.34.3-1.6.21
      pull_policy: IfNotPresent
      pull_secrets: []         # dedicated to the waiter Job; not borrowed from host.image.pull_secrets
    resources:
      limits:   { cpu: 100m, memory: 128Mi }
      requests: { cpu: 50m,  memory: 64Mi  }
```

The deprecated top-level `gke_autopilot_allowlist` key is still resolved via `coalesce` in `_labels.tpl`, so customers with existing values.yaml files keep working without changes.

## Test plan

- [x] `helm lint charts/shield -f charts/shield/ci/test-values.yaml` — passes
- [x] `helm unittest --strict -f 'tests/**/*_test.yaml' charts/shield` — **475 passed / 0 failed** (11 new waiter cases + 2 new daemonset coalesce cases + regression-guard assertions on the waiter Pod's allowlist label and the synchronizer's delete-policy)
- [x] Render matrix on `ci/test-values.yaml`:
  - default → 0 waiter resources
  - Autopilot, flag off → 0 waiter resources
  - generic, flag on → 0 waiter resources (outer guard wins)
  - Autopilot, flag on → all 4 waiter resources rendered, hook annotations correct
- [x] Deprecated `gke_autopilot_allowlist` key continues to drive the `cloud.google.com/matching-allowlist` label
- [x] `imagePullSecrets` block is omitted from the rendered Job when `gke_autopilot.allowlist_waiter.image.pull_secrets` is empty, and rendered correctly when populated
- [x] **End-to-end on a real GKE Autopilot cluster via FluxCD `HelmRelease`** — GKE 1.35.2-gke.1485000 / Flux 2.8.6:
  - **Negative control** (waiter disabled): host-shield DaemonSet manifest rejected by GKE Warden at admission time at `T+1.4s`, AllowlistSynchronizer reached Ready at `T+2s` — race reproduced exactly as theorized.
  - **Fix install** (waiter enabled): waiter Job ran, blocked on synchronizer Ready, exited 0; host-shield DaemonSet 4/4 ready, zero `FailedCreate` events, HelmRelease Ready=True.
  - **Upgrade path** (benign value change): pre-upgrade hook recycled the synchronizer (NotReady → Ready in 6 s), waiter blocked correctly, upgrade succeeded in ~22 s, non-disruptive (zero Pod restarts).

### Bugs caught and fixed during e2e

1. **Synchronizer hook missing `before-hook-creation` delete-policy.** Pre-upgrade would fail with `AlreadyExists`. Fixed pre-emptively (caught while planning the test). Unit-test assertion added.
2. **Waiter Pod self-blocked on Autopilot allowlist.** Pod template inherited `host.labels`, which carries `cloud.google.com/matching-allowlist`; GKE Warden rejected the waiter on the very allowlist it was waiting to load. Fixed by using `shield.labels` for the Pod template only (Job metadata still uses `host.labels` — Warden only checks Pods). Regression-guard assertion added. **This was unit-test invisible** — pure runtime admission-policy interaction.

## Checklist

- [x] Title of the PR starts with type and scope (`feat(shield):`)
- [x] Chart Version bumped (`1.36.1` → `1.37.0`)
- [x] Variables are documented in `README.md` (regenerated via `chart-doc-gen`) and `values.yaml`
- [x] All test files are in the `tests/` folder of the chart and have a `_test` suffix


charts/shield/values.yaml
```diff
@@ -645,5 +645,42 @@ fullname_override:
...
+    enabled: false
+    # Maximum time the Job will block on the AllowlistSynchronizer reaching Ready
+    timeout: 120s
+    # Override the name of the waiter ServiceAccount (defaults to <host fullname>-allowlist-waiter)
+    service_account_name:
+    image:
+      # The registry where the kubectl image is stored
+      registry: quay.io
+      # The repository where the kubectl image is stored
+      repository: sysdig/sysdig-kubectl
```
mavimo: Please use `sysdig/kubectl` as is the currently maintained version

```suggestion
      repository: sysdig/kubectl
```

charts/shield/values.yaml
```diff
@@ -645,5 +645,42 @@ fullname_override:
...
+      # The tag for the kubectl image (auto-bumped by .github/updatecli.d/config-update-kubectl-image.yaml)
+      tag: 1.34.1-1.5.17
```
mavimo:
```suggestion
      tag: 1.34.3-1.6.21
```

aroberts87 (approved):

PR #2653, opened 2026-06-11
feat(shield): add cluster.nats_password_existing_secret for external NATS password

aaronm-sysdig:
## What

New chart value `cluster.nats_password_existing_secret`. When set, Cluster
Shield uses the named Secret (key `sysdig-cluster-nats-password`) for the NATS
password instead of generating a random one, and mounts that Secret into the
container. Default `""` keeps the existing generate-a-random-password behaviour.

## Why

GitOps workflows that pre-render `helm template` and commit the manifests
(e.g. ArgoCD with raw manifests) get a new random password on every render,
which shows up as constant drift. An externally managed Secret keeps it stable.

## Backward compatibility

Purely additive. The default (`""`) is unchanged, so existing installs behave
exactly as before.

## Changes

- `values.yaml` / `values.schema.json` - new key, default `""`
- `templates/cluster/_secret.tpl` - skip generation when the value is set
- `templates/cluster/_helpers.tpl` - mount the named Secret when the value is set

## Testing

- `helm lint` clean
- Unset: cluster Secret has a generated password (unchanged)
- Value set: cluster Secret `data` is empty and the named Secret is mounted
- New unit tests added for the value and its mount; existing tests unchanged and green

### DIFF
--- a/charts/shield/templates/cluster/_helpers.tpl
+++ b/charts/shield/templates/cluster/_helpers.tpl
@@ -117,6 +117,7 @@ the inheritance logic lives in a single place.
     (include "common.credentials.access_key_secret_name" .)
     (include "common.credentials.secure_api_token_secret_name" .)
     (include "cluster.fullname" .)
+    .Values.cluster.nats_password_existing_secret
     (dig "cluster_scanner" "runtime_status_integrator" "nats_server" "password_existing_secret" nil .Values.cluster.additional_settings)
   -}}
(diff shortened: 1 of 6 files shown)


aaronm-sysdig:
Closing this - the underlying leak is already fixed in #2658 (chart 1.43.2), which strips the NATS password fields from the rendered config after the additional_settings merge. That covers the customer's case, so this is no longer needed.

PR #2738, opened 2026-09-11
feat(shield): grant RBAC for OpenShift Phase 2 resource collection

jaimeyh:

OpenShift posture collection currently covers two resources, `Route` and `DeploymentConfig`, each behind its own boolean. This PR extends it to nine more OpenShift resources by granting `get`/`list`/`watch` on each, so the KSPM collector can list them:

| `ocp_resources` value | apiGroup | resource |
| --- | --- | --- |
| `securitycontextconstraints` | `security.openshift.io` | `securitycontextconstraints` |
| `machineconfig` | `machineconfiguration.openshift.io` | `machineconfigs` |
| `machineconfigpool` | `machineconfiguration.openshift.io` | `machineconfigpools` |
| `machineset` | `machine.openshift.io` | `machinesets` |
| `buildconfig` | `build.openshift.io` | `buildconfigs` |
| `imagestream` | `image.openshift.io` | `imagestreams` |
| `template` | `template.openshift.io` | `templates` |
| `clusteroperator` | `config.openshift.io` | `clusteroperators` |
| `clusterresourcequota` | `quota.openshift.io` | `clusterresourcequotas` |

Rather than adding nine more booleans, resources are opted in through a single list setting:

```yaml
cluster:
  additional_settings:
    kspm_collector:
      ocp_resources:
        - securitycontextconstraints
        - buildconfig
```

A new helper, `cluster.posture_ocp_resource_enabled`, gates each rule on posture being enabled **and** the resource being listed, which is the same semantics the two existing flags already had. `ocp_route_enabled` and `ocp_deploymentconfig_enabled` keep working and now feed the same helper, so no current installation changes behaviour, and `route` / `deploymentconfig` can also be requested through `ocp_resources`.

This follows #2618, which added the RBAC for `Route` and `DeploymentConfig`, and mirrors the equivalent change in the `cluster-shield` chart.

Nothing is granted by default. With posture enabled and no `ocp_resources` set, the rendered ClusterRole is unchanged.

## Checklist

### DIFF
--- a/charts/shield/Chart.yaml
+++ b/charts/shield/Chart.yaml
@@ -13,5 +13,5 @@ maintainers:
   - name: mavimo
     email: marcovito.moscaritolo@sysdig.com
 type: application
-version: 1.49.5
+version: 1.49.6
 appVersion: "1.0.0"
--- a/charts/shield/templates/cluster/_config.tpl
+++ b/charts/shield/templates/cluster/_config.tpl
@@ -212,19 +212,37 @@
 {{- end }}
 
 {{/*
-  Checks if posture is enabled and OCP Route collection is enabled.
+  Checks if posture is enabled and the given OCP resource is listed in the
+  kspm_collector ocp_resources setting.
+  Takes a list of two elements: the chart context and the resource name.
+  Usage: include "cluster.posture_ocp_resource_enabled" (list . "route")
 */}}
-{{- define "cluster.posture_ocp_routes_enabled" -}}
-  {{- if and (include "cluster.posture_enabled" .) (dig "kspm_collector" "ocp_route_enabled" false .Values.cluster.additional_settings) -}}
+{{- define "cluster.posture_ocp_resource_enabled" -}}
+  {{- $ctx := index . 0 -}}
+  {{- $resource := index . 1 -}}
+  {{- $ocpResources := dig "kspm_collector" "ocp_resources" (list) $ctx.Values.cluster.additional_settings -}}
+  {{- if and (include "cluster.posture_enabled" $ctx) (has $resource ($ocpResources | default (list))) -}}
     {{- true -}}
   {{- end -}}
 {{- end }}
 
 {{/*
-  Checks if posture is enabled and OCP DeploymentConfig collection is enabled.
+  Checks if posture is enabled and OCP Route collection is enabled, either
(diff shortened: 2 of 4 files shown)


jaimeyh:
Parked on purpose, moving to draft. This is not waiting on review.

Per the discussion with @michele-mangili and @mavimo: chart changes become public as soon as they merge and customers can enable them right away, so the new configuration should not be published while the backend side is still being finished. Whether this eventually lands here or Siemens picks the chart from a branch is a product decision, and there is a call on Friday to settle the long-term shape. The likely outcome is no per-resource setting at all, with collection enabled for every OpenShift customer, in which case the gating in this PR gets simplified rather than merged as is.

Two things corrected before parking it, because a parked PR with a wrong constant is worse than an open one:

- Retargeted from `cluster-shield-release-1.27.0` to `cluster-shield-release-1.26.0`. The collector that reads `ocp_resources` ships in cluster shield **1.26.0**, confirmed by @michele-mangili. When this PR was opened `1.26.0-rc1` was already cut, so I had assumed 1.26.0 was closed.
- `cluster.posture_ocp_resources_min_version` corrected from `1.27.0-0` to `1.26.0-0` for the same reason. Left as it was, the rules would not have rendered on the exact release that carries the collector, and the failure is silent: the LIST returns 403 and `ocpNamespacedCollector.Collect` only logs `ocp collector failed` at Warn.

Test tags moved with the constant, including the two negative cases that use an older image, now pinned at `1.25.1` so they stay negative. `helm unittest --strict -f 'tests/*/*_test.yaml' ./charts/shield` passes, 37 suites and 570 tests.

What has to happen before this is picked up again:

1. The Friday call decides whether the per-resource setting survives.
2. If it does not, this PR shrinks to unconditional OpenShift rules gated on the image version only.
3. Either way `SSPROD-71653` has to be settled first. Four of the nine types, `clusteroperator`, `machineconfig`, `machineconfigpool` and `machineset`, cannot satisfy the collector's protobuf contract because `openshift/api` only generates protobuf for aggregated API server groups. They are accepted by the `ocp_resources` enum today and collect nothing, so any default-on decision would turn on four dead settings.

The equivalent change for the chart in `draios/cluster-shield` is draios/cluster-shield#1101, parked for the same reason.

